PR #19, opened 2019-08-05
GOV.UK Design System Version 3

peteryates:
Version 3 of the [GOV.UK Design System](https://design-system.service.gov.uk/) was recently released and it contains several breaking changes, outlined in their [changelog document](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md).

This PR aims to bring the form builder completely up to date with Version 3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,11 @@
 [![Test Coverage](https://api.codeclimate.com/v1/badges/fde73b5dc9476197281b/test_coverage)](https://codeclimate.com/github/DFE-Digital/govuk_design_system_formbuilder/test_coverage)
 
 This gem provides a easy-to-use form builder that generates forms that are
-fully-compliant with the [GOV.UK Design System](https://design-system.service.gov.uk/),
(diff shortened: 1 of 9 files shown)


paulrobertlloyd:
Out of interest, would this change make this gem incompatible with sites still using V2 of the design system?

peteryates:
@paulrobertlloyd it will, yes.

I am currently deciding on how to progress, if there is a call to keep the `0.7` releases going, supporting GOV.UK Design System V2 with minor fixes and tweaks, I'd be happy to do that. As far as I can tell, the only people using it are already targetting V3, but I'd be happy to hear otherwise 🙂

Moving to GOV.UK Design System V3 seems like a key part of the path towards `1.0`.

paulrobertlloyd:
@peteryates All sounds good to me! Should support for V2 be needed by someone down the line, that can always be added later.

paulrobertlloyd:
I guess the only thing to add is that it might be worth calling out this as a requirement in the read me.

PR #20, opened 2019-08-06
Display one error message per field in error summary

RobertM905:
### Context

The error summary displayed all field errors grouped on one line.

This commit changes the error summary to only display the first error for a field.

This PR is a draft as we would like to discuss the testing approach.

### Changes proposed in this pull request

Before
Screenshot 2019-08-06 at 14 37 38

----

After
Screenshot 2019-08-06 at 14 35 06

### Guidance to review

To test this feature a model with two errors on one field was needed. 

We added an extra validation to the existing Person class.

We create an invalid person for these tests and then expect to see the correct message in the error summary. 

This feels brittle and we would like to discuss it.

### DIFF
--- a/spec/govuk_design_system_formbuilder/builder/error_summary_spec.rb
+++ b/spec/govuk_design_system_formbuilder/builder/error_summary_spec.rb
@@ -31,18 +31,34 @@ describe GOVUKDesignSystemFormBuilder::FormBuilder do
         )
       end
 
-      specify 'the error summary should contain a list with all the errors included' do
-        expect(subject).to have_tag('ul', with: { class: %w(govuk-list govuk-error-summary__list) }) do
-          expect(subject).to have_tag('li', count: object.errors.count)
-        end
-      end
-
-      context 'error messages' do
+      describe 'error messages' do
         subject! { builder.send(method) }
 
+        context 'there are multiple errors each with one error message' do
+          let(:object) { Person.new(favourite_colour: nil, projects: []) }
+
+          specify 'the error summary should contain a list with one error message per field' do
+            expect(subject).to have_tag('ul', with: { class: %w(govuk-list govuk-error-summary__list) }) do
+              expect(subject).to have_tag('li', text: 'Choose a favourite colour')
+              expect(subject).to have_tag('li', text: 'Select at least one project')
+            end
+          end
+        end
+
+        context 'there are multiple errors and one has multiple error messages' do
+          let(:object) { Person.new(name: nil, favourite_colour: nil) }
+
+          specify 'the error summary should contain a list with one error message per field' do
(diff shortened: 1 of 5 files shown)


spec/govuk_design_system_formbuilder/builder/error_summary_spec.rb
```diff
@@ -31,18 +31,48 @@
...
       end
 
-      specify 'the error summary should contain a list with all the errors included' do
-        expect(subject).to have_tag('ul', with: { class: %w(govuk-list govuk-error-summary__list) }) do
-          expect(subject).to have_tag('li', count: object.errors.count)
+      describe 'error messages' do
+        subject! { builder.send(method) }
+
+        context 'there are multiple errors each with one error message' do
+          let(:object) do
```
peteryates: I think this can be expressed a little more concisely

```ruby
let(:object) { Person.new(favourite_colour: nil, projects: []) }
```

RobertM905:
We could also have a look at how error messages are displayed next to the form field inputs and add them to this PR?

peteryates:
> We could also have a look at how error messages are displayed next to the form field inputs and add them to this PR?

If you think it fits in with this one, sure thing. Really don't mind if they're combined or separate

peteryates:
Looks good. I think that this behaviour is probably worth covering with a spec There's a shared context called `'a field that supports errors` that is used by all of the error message including specs.

Something like this might work, welcome to hear your thoughts on the approach

```diff
index 4ee4ba0..258932d 100644
--- a/spec/support/shared/shared_error_examples.rb
+++ b/spec/support/shared/shared_error_examples.rb
@@ -19,6 +19,26 @@ shared_examples 'a field that supports errors' do
         id: error_identifier
       })
     end
+
+    context 'when there is more than one error' do
+      let(:first_error) { "It is totally wrong" }
+      let(:second_error) { "Very wrong indeed" }
+      let(:error_messages) { [first_error, second_error] }
+
+      before do
+        allow(object.errors.messages).to(
+          receive(:[]).and_return(error_messages)
+        )
+      end
+
+      specify 'the first error should be included' do
+        expect(subject).to have_tag('span', text: Regexp.new(first_error))
+      end
+
+      specify 'the second error should not be included' do
+        expect(subject).not_to have_tag('span', text: Regexp.new(second_error))
+      end
+    end
   end
 
   context 'when the attribute has no errors' do
```

RobertM905:
We did look into writing a test to cover this but we couldn't decide on a nice way to stub `object.errors.messages`.
This is a much cleaner way Cheers. 👍

peteryates:
Great, thanks.

With regards to the difficult you found in stubbing `object.errors.message`, is there anything about the way the specs are set up that could be improved? I tried to keep it generic and re-usable but perhaps that was at the expense of it being a little _opaque_?

I'm happy with these changes, I'll merge shortly 👍🏽

PR #57, opened 2019-10-31
Allow date input to capture only month and year

### LINKED ISSUE #56: Date fields without day field
Hello! Our service has date fields that look like this:

Screenshot 2019-10-30 at 11 28 43

It would be nice if the [date field component](https://govuk-form-builder.netlify.com/form-elements/date-field/) allowed this pattern.

Currently we still use it but we hide the day field using `display: none` in CSS. This has the downside of breaking the link from the error summary (because it points to the day field).

peteryates:
[Sometimes non-exact inputs are required](https://github.com/DFE-Digital/apply-for-postgraduate-teacher-training/pull/404) for dates that the user might not remember, such as their graduation date or when they started a new job. In this case, don't ask for the day - only the month and year.

I believe that Rails will automatically set the date to the first of the month so this approach _should just work_.

Screenshot 2019-10-31 at 20 17 34

* [x] Amend the date input so it can be rendered sans day
* [x] Check that the day-less input works ok with Rails and saves valid dates
* [x] Come up with a better name for the param than `smallest_segment` 🍰
* [x] Add some specs covering the behaviour of the error summary and linking directly to the month field when date is omitted

cc @chubberlisk @tvararu 

Fixes #56

### DIFF
--- a/guide/content/form-elements/date-field.slim
+++ b/guide/content/form-elements/date-field.slim
@@ -41,6 +41,6 @@ section
       | When asking for dates that the user might not know exactly, such as
         their graduation date or their start date at a past job, the exact date
         might not be necessary. The day can be optionally omitted by providing
-        the <code>smallest_segment</code> parameter.
+        the <code>omit_day</code> parameter.
 
 == render('/partials/related-info.*', links: date_info)
--- a/guide/content/form-elements/date-field.slim
+++ b/guide/content/form-elements/date-field.slim
@@ -23,4 +23,24 @@ section
     caption: "Date field with legend and hint and birth date autocompletion",
     code: date_field)
 
+  == render('/partials/example-fig.*',
+    caption: "Month field",
+    code: month_field,
+    hide_html_output: true) do
+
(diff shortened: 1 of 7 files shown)


peteryates:
Yeah, I mentioned in 660690f that capturing just a year is probably better served by a regular `govuk_number_field` with it stored as an integer rather than a date with day and month defaulting to `01`.

I had a look for previous discussions about using this pattern to capture only month and year but couldn't find anything (hence the warning in the guide). Do you have any links to previous discussions?

I'm leaning more towards `:omit_day` now. I'll update it to reflect that.

Also, can you confirm Rails' behaviour when a date is submitted with no day, does it always default it to `01`?

tvararu:
> Do you have any links to previous discussions?

None of the top!

> Also, can you confirm Rails' behaviour when a date is submitted with no day, does it always default it to 01?

```ruby
pry(main)> Date.new(2019, 5)
=> Wed, 01 May 2019
```

Seems okay 👍

PR #86, opened 2019-12-26
Bump Ruby and Rails versions

peteryates:
Bump Ruby and Rails to latest versions

### Ruby

* `2.5.7`
* `2.6.5`
* `2.7.0`

### Rails
* `5.2.4`
* `6.0.2`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ language: ruby
 # cache: bundler
 
 rvm:
-  - 2.5.6
-  - 2.6.4
+  - 2.5.7
+  - 2.6.5
+  - 2.7.0
 
 script:
   - make ruby-lint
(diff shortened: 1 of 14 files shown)


peteryates:
Everything passes but `2.7.0` outputs _a lot_ of warnings due to [this change](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/)

```
spec/govuk_design_system_formbuilder/builder/configuration_spec.rb:15: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

Rails also seems to output a load of warnings, the I'm ignoring them for the time being using this kludgey filter

```bash
be rspec spec/govuk_design_system_formbuilder/builder/configuration_spec.rb 2>&1 | ag -v 'i18n|active_model|action_view'
```

PR #90, opened 2020-01-07
Allow plain ruby objects

peteryates:
Allow the form builder to operate with plain Ruby objects in addition to `ActiveModel` and `ActiveRecord` ones that support `#errors`. This is useful when performing transient operations like searching.


peteryates:
@jsugarman, would you mind giving this branch a spin and if it's all good I'll merge and get a mini release (with this and #89) prepped. Thanks

Issue #88, opened 2020-01-06
Enable use for plain ruby objects


jsugarman:
Hi @peteryates, this is great stuff and was looking to implement in a product I am working on over in MoJ. 

One problem is you cannot use it against a plain ruby class due to the lack of `errors` on the object. I am just looking to create a basic search form with options.

I could monkey patch this or create a dummy `errors` method to patch this app-side but think it could be beneficial generally. I have a branch that I have it working in (with tests) but cannot push it - receive a `403`.

peteryates:
Thanks Joel. I agree that adding the functionality to the gem is _probably_ a good idea - I can't see any reason not to! If you [fork this repo and create a PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork) I'll gladly review your changes and collaborate on adding the feature.

peteryates:
I had a _quick_ attempt @jsugarman, will this (#90) work for you?

PR #91, opened 2020-01-07
Enable use for plain ruby classes

jsugarman:
#### What
Allow creation of GDS design system form components for plain ruby classes. [Issue](https://github.com/DFE-Digital/govuk_design_system_formbuilder/issues/88).

#### Why
Inline with standard rails form builder helpers and to facilitate creation of simple forms with no models associated.

For example to create a simple search form it would be nice to not have to create any model and just do something like 

```
# routes.rb
post 'search', action: :index, controller: :search
```

```
# app/controllers/search_controller.rb
class SearchController < ApplicationController
  def index
    @search_options = [[:opt1, 'Option 1'], [:opt2, 'Option 2']]
  end
end
```

```
# app/view/search/index.html.haml
= form_with url: search_path, local: true do |f|
  = f.govuk_collection_radio_buttons :search_option,
      @search_options,
      :first,
      :last,
      legend: { text: 'Choose a filter' },
      hint_text: 'Specify a filter for the search'
  = f.govuk_text_field(:query, label: { text: 'Find a thing' }, hint_text: 'enter thing to find' )
  = f.govuk_submit('Search')

```

Currently this is only prevented by the use of calls to `errors`, from what i can see.

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/error_summary.rb
+++ b/lib/govuk_design_system_formbuilder/elements/error_summary.rb
@@ -10,7 +10,7 @@ module GOVUKDesignSystemFormBuilder
       end
 
       def html
-        return nil unless object_has_errors?
+        return unless object_has_errors?
 
         content_tag('div', class: summary_class, **error_summary_attributes) do
           safe_join(
@@ -65,6 +65,8 @@ module GOVUKDesignSystemFormBuilder
       end
 
       def object_has_errors?
+        return unless @builder.object.respond_to?(:errors)
+
         @builder.object.errors.any?
       end
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -57,9 +57,9 @@ RSpec.configure do |config|
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
 
-# The settings below are suggested to provide a good initial experience
-# with RSpec, but feel free to customize to your heart's content.
-=begin
+  # The settings below are suggested to provide a good initial experience
+  # with RSpec, but feel free to customize to your heart's content.
+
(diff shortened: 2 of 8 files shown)


lib/govuk_design_system_formbuilder/elements/error_summary.rb
```diff
@@ -65,6 +65,8 @@ def error_summary_title_id
       end
 
       def object_has_errors?
+        return unless @builder.object.respond_to?(:errors)
```
peteryates: I think the builder _should_ error if the dev tries to render an error summary in a form with an object that doesn't support errors

spec/spec_helper.rb
```diff
@@ -57,9 +57,9 @@
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
 
-# The settings below are suggested to provide a good initial experience
-# with RSpec, but feel free to customize to your heart's content.
-=begin
+  # The settings below are suggested to provide a good initial experience
```
peteryates: Happy for this to be amended but would prefer a separate dedicated PR

jsugarman: no worries. will push one

lib/govuk_design_system_formbuilder/elements/error_summary.rb
jsugarman: > Overall this looks good. Before this was raised I'd already created #90 which does more or less the same thing. If you're happy for me to I'll cherry pick the tests from this PR into that one and tweak accordingly. Thanks @jsugarman

Yep, what ever you think is best.

jsugarman: > I think the builder _should_ error if the dev tries to render an error summary in a form with an object that doesn't support errors

👍 Makes sense, better to catch these things early.

PR #96, opened 2020-01-29
Add classes option to checkboxes collection component

tvararu:
The Design System allows you to pass in any number of CSS classes, it feels like we should do this as well.


peteryates:
This looks good! I'll have a closer look tomorrow, thank you for the PR :+1:

PR #105, opened 2020-03-30
Don’t wrap submit buttons in govuk-form-group

duncanjbrown:
Following discussion on DfE Slack, don't wrap `submit` elements in an unnecessary `govuk-form-group`.

### DIFF
--- a/spec/govuk_design_system_formbuilder/builder/submit_spec.rb
+++ b/spec/govuk_design_system_formbuilder/builder/submit_spec.rb
@@ -7,10 +7,8 @@ describe GOVUKDesignSystemFormBuilder::FormBuilder do
     let(:args) { [method] }
     subject { builder.send(method, text) }
 
-    specify 'output should be a form group containing a submit input' do
-      expect(subject).to have_tag('div', with: { class: 'govuk-form-group' }) do |fg|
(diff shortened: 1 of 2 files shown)


spec/govuk_design_system_formbuilder/builder/submit_spec.rb
```diff
@@ -7,10 +7,8 @@
     let(:args) { [method] }
     subject { builder.send(method, text) }
 
-    specify 'output should be a form group containing a submit input' do
-      expect(subject).to have_tag('div', with: { class: 'govuk-form-group' }) do |fg|
-        expect(fg).to have_tag('input', with: { type: 'submit' })
-      end
+    specify 'output should be a submit input' do
+      expect(fg).to have_tag('input', with: { type: 'submit' })
```
peteryates: Think this'll fail now `fg` isn't declared by the removed expectation block:

```suggestion
      expect(subject).to have_tag('input', with: { type: 'submit' })
```

duncanjbrown: Oops — my mistake. Should be better now?

peteryates: :+1:

PR #111, opened 2020-04-02
Reorder the elements in govuk_date_field

### LINKED ISSUE #110: Can't add content between h1 and hint text
Example intended design:

image

We also have other screens where there is longer text between the question (which is both legend and h1) and hint text. This design is acceptable by the design system but when we inject content it is in the wrong position:

image

and shown with error:

image

peteryates:
The injected content being rendered immediately above the inputs didn't make sense and it increased the distance between the hint/errors and input elements.

The new order is:

0. fieldset legend
1. injected content
2. hint
3. error (if invalid)
4. date part inputs

Fixes #110


peteryates:
cc @simcast - thanks for reporting, I _think_ this fixes it and the new order definitely makes more sense :+1:

zheileman:
@peteryates Hi Peter, thank you. Our service designer raised this problem.

We've seem this also happens in other elements, not only dates. Do you think it would be best to make it consistent for any element that allows supplemental content?

Thank you.

PR #123, opened 2020-05-12
Extend legend options to enable visually hidden

zheileman:
Allow to pass an optional option to the existing legend options hash, in a similar way to the one in labels, to add an extra class `govuk-visually-hidden` to the legend for those cases where we want to avoid visual repetition.

This is in line with recommendations from the design system:
https://design-system.service.gov.uk/get-started/labels-legends-headings/

However, this option only removes visual duplication and will not help users of screen readers. They will still hear both the page heading and the visually hidden `<label>` or `<legend>`.

To prevent this, set the contents of the `<label>` or `<legend>` as the page heading.

### DIFF
--- a/guide/content/introduction/labels-hints-and-legends.slim
+++ b/guide/content/introduction/labels-hints-and-legends.slim
@@ -130,6 +130,15 @@ section
         | The legend size which follows the usual GOV.UK pattern of
           <code>xl</code>, <code>l</code>, <code>m</code> and <code>s</code>
 
+    div.govuk-summary-list__row
+      dt.govuk-summary-list__key
+        code hidden
+
+      dd.govuk-summary-list__value
+        | Legends can be visually hidden when their presence would complicate
+          the form. The hidden value defaults to <code>false</code> and can be
+          overridden with <code>true</code>
+
   == render('/partials/example-fig.*',
     caption: "Radio buttons with a fully-customised legend",
     code: radios_with_legend,
--- a/lib/govuk_design_system_formbuilder/builder.rb
+++ b/lib/govuk_design_system_formbuilder/builder.rb
@@ -315,6 +315,7 @@ module GOVUKDesignSystemFormBuilder
     # @option legend text [String] the fieldset legend's text content
     # @option legend size [String] the size of the fieldset legend font, can be +xl+, +l+, +m+ or +s+
     # @option legend tag [Symbol,String] the tag used for the fieldset's header, defaults to +h1+, defaults to +h1+
+    # @option legend hidden [Boolean] control the visibility of the legend. Hidden legends will still be read by screenreaders
     # @return [ActiveSupport::SafeBuffer] HTML output
     #
     # @example A collection of radio buttons for favourite colours, labels capitalised via a proc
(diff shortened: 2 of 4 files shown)


zheileman:
Hello 👋 
I will be contributing a few improvements or features we needed for one of our projects.
Hopefully these are also useful to others.

peteryates:
Hi, thanks for this.

Just to clarify, on some pages you want a header at the top of the page followed by a form that contains a legend, and that legend might have the same text as the heading above, so this hides the fieldset heading?

My understanding of the [Making labels and legends headings](https://design-system.service.gov.uk/get-started/labels-legends-headings/) design is that legend would _become_ the heading, a behaviour that's already supported by the builder for both [labels](https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder/Builder#govuk_text_field-instance_method) and [legends](https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder/Builder#govuk_collection_radio_buttons-instance_method) via their `tag` argument:

>Options Hash (label:):
> * text (String) — the label text
> * size (String) — the size of the label font, can be `xl`, `l`, `m`, `s` or `nil`
>  * tag (Symbol, String) — the label's wrapper tag, intended to allow labels to act as page headings
> * hidden (Boolean) — control the visability of the label. Hidden labels will still be read by screenreaders

> Options Hash (legend:):
>  * text (String) — the fieldset legend's text content
>  * size (String) — the size of the fieldset legend font, can be `xl`, `l`, `m` or `s`
>  * tag (Symbol, String) — the tag used for the fieldset's header, defaults to `h1`

The standalone heading wouldn't be needed and repetition would be avoided while remaining in line with the suggested approach in the design system docs.

zheileman:
Hi @peteryates thanks for your reply.

Sorry if it was not very clear. Basically, in some places in our service we have the following that it was not possible out of the box with the gem, unless I'm missing something:

Screen Shot 2020-05-12 at 14 39 59

Where in order to make it more accessible to screen readers, we wanted the checkbox legend to be hidden as you can see in the markup it produces:

Screen Shot 2020-05-12 at 14 40 31


Also, although not ideal, we have some complex and lengthy copy-wise pages that make the radios very disjointed from the heading. I know we can include the heading as part of the legend so we don't need to make it hidden, that's what we've done in the majority of pages, but for these long pages it worked best for us in this way:

Screen Shot 2020-05-12 at 14 41 49

And this is the markup it produces for screen readers:

Screen Shot 2020-05-12 at 14 42 12

So we needed the possibility to make the legends hidden, same as it was already possible with the labels.

peteryates:
Ah, that makes more sense. Thanks for clarifying.

zheileman:
Thanks @peteryates, I'm running rubocop locally and it is not picking any offences. Not sure why on Travis it fail. I just run on my local branch `bundle exec rubocop` is there anything I'm missing?

peteryates:
Yeah - it's my fault! Until [this PR](https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/122/files), which was merged a few minutes ago the exact version of Rubocop wasn't specified in the `gemspec` and a couple of (mostly incorrect!) cops were added between 3.7 and 3.9 that caused failures.

As you can see, I'm forcing it to remain at 3.9 for the time being.

zheileman:
Ah cool, thanks. I will rebase my PR to include those changes and make sure I've not introduced any offence 👍

PR #125, opened 2020-05-12
Improve localisation

zheileman:
> ⚠️ Please note I've tried to maintain backwards compatibility as much as possible, and no method signatures are changed. However, it would be great to run this branch against a consolidated project, to see if the changes in localisation keys mess around with something. We can come up with any fixes if that happens.
Our service is heavy on localisation from simple Yes/No radios, to hundreds of check boxes and more complex radio options, and these are the changes we've been using.

Extend the localisation schema with the `value` of the radio/checkbox option (for example "yes" or "no") to be able to localise these automatically.

This change should make it now possible to set separately labels and/or hints for either the fieldset or the individual radios/checkboxes inside that fieldset.

In order to not break the current interface, the `#govuk_collection_radio_buttons` method does not change, which means to be able to take the label from the locales, instead of providing to the method a `text_method`, just pass `nil`.

Example:

```ruby
f.govuk_collection_radio_buttons :parent, @values, :value, nil, inline: true
```

This makes for a bit ugly interface but could be improved later on. I didn't want to break compatibility.

Also, now localisation can also contain html elements, and in order to follow the usual Rails conventions, these locale keys will contain the suffix `_html` which will make the content html safe automatically when using `t()` method in the view, although it does not happen automatically outside the view that's why I use `#html_safe` if a locale was found.

Still, the non-html key will be tried first, and returned if one is found, otherwise the html version will be tried, and if none was found, nil will be returned as before.

Additionally I've done a small refactor so the collection check boxes hint elements are the same as with other form elements, by using the same class and method signature.

I couldn't find a reason why this was not the case, probably it was a leftover from a previous refactor.


peteryates:
Thanks, this looks positive.

I did go some way to addressing this by supporting [setting labels, hints and values with procs](https://govuk-form-builder.netlify.app/form-elements/radios/#populating-values-labels-and-hints-with-procs) but this looks clearer; checkboxes having their own mini builder object in Rails made it clunky.

I'll have a closer look, I think there might be room for implementing this without requiring the `nil` argument - I agree that's not idea.

lib/govuk_design_system_formbuilder/traits/localisation.rb
```diff
@@ -4,28 +4,39 @@ module Localisation
...
+      def schema_path
+        [].tap do |path|
+          path.push(@object_name)
+
+          if @value.present?
+            path.push("#{@attribute_name}_options", @value)
+          else
+            path.push(@attribute_name)
+          end
+        end
```
peteryates, line 37: I think this can be simplified a little further

```suggestion
        if @value.present?
          [@object_name, "#{@attribute_name}_options", @value]
        else
          [@object_name, @attribute_name]
        end
```

lib/govuk_design_system_formbuilder/traits/localisation.rb
```diff
@@ -4,28 +4,39 @@ module Localisation
...
-        key = localisation_key(context)
+        return unless @object_name.present? && @attribute_name.present?
 
-        # `I18n.exists?(nil)` returns true when `config.i18n.fallbacks` is
-        # enabled, so only proceed if the key is present too
-        return nil unless key.present? && I18n.exists?(key)
-
-        I18n.translate(key)
+        I18n.translate(localisation_key(context), default: nil) ||
+          I18n.translate(localisation_key(context, html: true), default: nil).try(:html_safe)
```
peteryates, line 10: I feel that this might be a bit much for a single line, perhaps breaking it up might make it easier to grasp?

```suggestion
        localise(context) || localise_html(context)
      end

      def localise(context)
        I18n.translate(localisation_key(context), default: nil)
      end

      def localise_html(context)
        I18n.translate(localisation_key(context, html: true), default: nil).try(:html_safe)
      end
```

lib/govuk_design_system_formbuilder/traits/localisation.rb
zheileman: Yeah, totally. This was part of an earlier version as I was playing around with the localisation and then I forgot to cleanup the unnecessarily complex code. Thanks for pointing this out!

lib/govuk_design_system_formbuilder/traits/localisation.rb
zheileman: I agree. Will make the amendments and push!

zheileman:
Thanks! Will make the suggested changes and also try to come up with some documentation for the guide. Not very familiar with the documentation framework but will try to take other sections as examples 👍

peteryates:
I'm more than happy to assist with that, or we can get this PR in and I'll create an issue to document it separately :+1: 

With regards to running the guide I tend to run the `make keep-building-guide` and `make watch-guide` in two tmux panes, then you should be able to preview your changes at `http://localhost:3006`.

You'll need `entr` installed to run the watch-guide command.

I aim to make it work with [`nanoc-live`](https://rubygems.org/gems/nanoc-live/versions/1.0.0a1) but I can't see how to use groups in a `gemspec` and it needs to be in the `:nanoc` group :man_shrugging:

zheileman:
Thanks! got the documentation up and running locally and added some new examples for the locales, taking "inspiration" from the existing examples 😉 

Feel free to change or add anything else you like or let me know and will do it. The build is failing but it looks like some temporary issue with Travis.

zheileman:
Thank you! 🎉

PR #127, opened 2020-05-14
Support other design systems

peteryates:
A team from the NHS working on COVID19 response projects reached out, they would like to use this form builder to aid their work.

This is an exploratory branch testing out the feasibility of the form builder supporting other compatible design systems, namely [that used by the NHS](https://service-manual.nhs.uk/design-system).

By compatible, I'm referring to design systems where the form elements match the GOV.UK version in all but the prefixes of the class names.

In the NHS version, it appears that `govuk` has just been replaced with `nhsuk`, so a label would [look like this](https://service-manual.nhs.uk/design-system/components/text-input):

```html
<label class="nhsuk-label">First name</label>
```

## How it works

There is a new configuration option called `brand` that defaults to `govuk`.

All hard-coded references to `govuk` throughout the form builder have been replaced to calls to the `#brand` method which in turn pulls the value from the config.

## What's left to do

* [x] Await feedback from the NHS team to ensure everything's working as they expect
* [x] Tests, tests, tests
* [x] Version bump, to `1.2.0` (beta pushed, `1.2.0b1`)
* [x] Updating the guide and docs where appropriate
* [x] Explore the possibility of aliasing the form builder helper methods if a non-default `brand` is used, eg `= form.nhsuk_text_field(...)`
* [x] Update the `README` accordingly

## Preview
Screenshot from 2020-05-15 17-00-58


WorkSutton:
@peteryates are you able to get in contact with either myself (Mark Sutton) or Nuno Costa who are both working on the PHE project who you have adapted this for?
Mark.X.Sutton@phe.gov.uk

peteryates:
It [appears](https://github.com/nhsuk/nhsuk-frontend/issues/608) that there are some _deeper_ differences between the NHS Design System and the GOV.UK Design System than initially thought. Some JavaScript functionality in the NHS DS isn't properly triggered because of slightly-differing class names/attributes.

The NHS team using the form builder have had to monkey patch some methods to bring them in line.

This gem will not claim that it supports the NHS DS, but the work in this PR is still valid and could be used as the basis for a companion gem that applies refinements/monkey patches where required to add the necessary markup needed by the NHS DS; it would also need its own set of tests.

* [ ] Make it obvious that the NHS Design System isn't officially supported
* [ ] Set the branding guide page to `draft` but leave the API documentation in place

PR #128, opened 2020-05-15
Add custom classes to the submit element

cpjmcquillan:
## In this PR
- Allow a custom class to be passed as a string to the submit element (e.g `= f.govuk_submit 'Create', custom_class: 'custom-govuk-class'`
- Append the custom class to the built element
- Add a basic spec to check the custom class is added
- Add a section to the guide

### DIFF
--- a/lib/govuk_design_system_formbuilder/builder.rb
+++ b/lib/govuk_design_system_formbuilder/builder.rb
@@ -567,6 +567,7 @@ module GOVUKDesignSystemFormBuilder
     # @param secondary [Boolean] makes the button grey ({https://design-system.service.gov.uk/components/button/#secondary-buttons secondary}) when true
     # @todo The GOV.UK design system also supports {https://design-system.service.gov.uk/components/button/#disabled-buttons disabled buttons}, they
     #   should probably be supported too
+    # @param classes [String] Classes to add to the submit button
     # @param prevent_double_click [Boolean] adds JavaScript to safeguard the
     #   form from being submitted more than once
     # @param validate [Boolean] adds the formnovalidate to the submit button when true, this disables all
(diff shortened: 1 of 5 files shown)


lib/govuk_design_system_formbuilder/builder.rb
```diff
@@ -586,8 +586,8 @@ def govuk_check_box(attribute_name, value, hint_text: nil, label: {}, link_error
     #   = f.govuk_submit "Proceed", prevent_double_click: true do
     #     = link_to 'Cancel', some_other_path, class: 'govuk-button__secondary'
     #
-    def govuk_submit(text = config.default_submit_button_text, warning: false, secondary: false, prevent_double_click: true, validate: false, &block)
-      Elements::Submit.new(self, text, warning: warning, secondary: secondary, prevent_double_click: prevent_double_click, validate: validate, &block).html
+    def govuk_submit(text = config.default_submit_button_text, warning: false, secondary: false, custom_class: nil, prevent_double_click: true, validate: false, &block)
```
peteryates, line 589: Can we have a line of explanation for `custom_class` above this method for the rdoc please?

lib/govuk_design_system_formbuilder/builder.rb
```diff
@@ -586,8 +586,8 @@ def govuk_check_box(attribute_name, value, hint_text: nil, label: {}, link_error
...
+      Elements::Submit.new(self, text, warning: warning, secondary: secondary, custom_class: custom_class, prevent_double_click: prevent_double_click, validate: validate, &block).html
```
peteryates, line 590: [Some other helpers](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/builder.rb#L392) in the gem already contain this functionality via the `classes` param. I think this should be called `classes` too for consistency.

PR #130, opened 2020-05-19
Fix block handling bug on #govuk_fieldset

peteryates:
The legend wasn't being displayed by `Containers::Fieldset` when called directly (rather than via another helper), this was due to some weirdness calling `yield` inside `safe_join` in the context of Rails templates.

It's only apparent in `#govuk_fieldset` because it's the only Container class to have its own helper method. `Container` classes work differently to `Element` ones in that they take their block via the HTML method rather than initialize.

There is already more-robust block handling built into the base class for elements, now the `Fieldset` can be called with either a block passed into the initializer or with one passed into `#html`.


aldavidson:
I can confirm this fixes the issue on my machine - thanks!

PR #144, opened 2020-06-08
Disabled submit buttons

cpjmcquillan:
Strictly speaking, HTML requires just the `disabled` attribute to be present on the input element, but if it is present with a value it will have the same effect.

## In this PR
- Allow disabled attribute to be passed to the submit element
- Add disabled class to disabled elements
- Update guide

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/submit.rb
+++ b/lib/govuk_design_system_formbuilder/elements/submit.rb
@@ -3,7 +3,7 @@ module GOVUKDesignSystemFormBuilder
     class Submit < Base
       using PrefixableArray
 
-      def initialize(builder, text, warning:, secondary:, classes:, prevent_double_click:, validate:, &block)
+      def initialize(builder, text, warning:, secondary:, classes:, prevent_double_click:, validate:, disabled:, &block)
         fail ArgumentError, 'buttons can be warning or secondary' if warning && secondary
 
         @builder              = builder
@@ -13,6 +13,7 @@ module GOVUKDesignSystemFormBuilder
         @secondary            = secondary
         @classes              = classes
         @validate             = validate
+        @disabled             = disabled
         @block_content        = capture { block.call } if block_given?
       end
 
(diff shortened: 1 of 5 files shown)


lib/govuk_design_system_formbuilder/elements/submit.rb
```diff
@@ -48,13 +50,18 @@ def padding_class(content_present)
...
+
       def extra_args
+        disabled_hash = @disabled ? { disabled: 'disabled' } : {}
         {
           formnovalidate: !@validate,
           data: {
             module: %(#{brand}-button), 'prevent-double-click' => @prevent_double_click
           }.select { |_k, v| v.present? }
-        }
+        }.merge!(disabled_hash)
```
peteryates, line 64: Can this be simplified slightly? I _think_ this will work in the same way:

```suggestion
        {
          formnovalidate: !@validate,
          disabled: @disabled,
          data: {
            module: %(#{brand}-button), 'prevent-double-click' => @prevent_double_click
          }.select { |_k, v| v.present? }
        }
```

cpjmcquillan: My understanding is that disabled attribute with **any** value disables the input, which is why I did it this way. https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute

peteryates: Yeah, you're absolutely right with regards to the HTML, but Rails' tag helper [will discard](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/tag_helper.rb#L77) any [boolean attributes](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/tag_helper.rb#L16) that aren't truthy.

I've done a quick test with the above modification and it appears to work work in the same manner.

The one _danger_ is that someone might try to disable it with `disabled: 'false'` which wouldn't have the desired effect, but that would be in line with Rails' default behaviour (I tested that too).

PR #149, opened 2020-06-26
Fix caption leaking its localisation to radio options

zheileman:
We run into a problem when using the new [captions capability](https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/135) of the gem where the localised caption of `govuk_collection_radio_buttons` or `govuk_radio_buttons_fieldset` will leak also to each of their corresponding radio option labels.

I think there is not a use-case to use captions in the labels of individual radio or check box options and couldn't find anything in the design system documentation about this, so I opted for the simplest possible solution, but if in the future there is a use-case for having captions also in these option labels, then I suppose a bit of work is neccessary in the localisation to not inherit that of the fieldset.

I didn't see this happening to check boxes, at least in the scenarios I have, but just in case I'm also excluding these check boxes from being "captionable".

Before:
Screen Shot 2020-06-26 at 12 48 57

After:
Screen Shot 2020-06-26 at 12 53 18

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/label.rb
+++ b/lib/govuk_design_system_formbuilder/elements/label.rb
@@ -39,7 +39,7 @@ module GOVUKDesignSystemFormBuilder
 
       def label
         @builder.label(@attribute_name, label_options) do
-          @content || safe_join([caption_html, @text])
+          @content || safe_join([caption, @text])
         end
       end
 
@@ -61,8 +61,8 @@ module GOVUKDesignSystemFormBuilder
         }
       end
 
-      def caption_html
-        caption_element.html if [@radio_class, @checkbox_class].all?(nil)
+      def caption
+        caption_element.html unless [@radio_class, @checkbox_class].any?
       end
 
       def radio_class(radio)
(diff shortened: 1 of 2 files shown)


lib/govuk_design_system_formbuilder/elements/label.rb
```diff
@@ -61,6 +61,10 @@ def label_options
         }
       end
 
+      def caption_html
```
peteryates, line 64: I think to be consistent with [other similar methods](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/containers/fieldset.rb#L40) this could just be `caption`

zheileman: Yep! I will change it.

zheileman:
Hi Pete! thanks for the feedback!

I get what you are suggesting but I believe that could be part of a bigger refactor, and for consistency maybe also be done in the `elements/hint.rb` where the the same mechanics are used: `radio_class(radio)` and `checkbox_class(checkbox)`.

Didn't want to do it as part of this work as seemed a bit unrelated and more prone to introduce regressions, so I settled with this simpler solution of using the classes so I didn't have to pass around the `radio` or `checkbox` booleans.

PR #160, opened 2020-07-07
Form group custom classes

### LINKED ISSUE #158: Allow customisation of form group
As mentioned in the comments on #153 it could be useful to allow setting classes on the `<div class="govuk-form-group">` element that wraps most form controls (everything except `govuk_submit` since #105).

This should work consistently across all the builder methods that construct a form group. The most straightforward approach would be to simply add a `form_group_classes:` keyword arg and ensure it's passed into [`FormGroup`](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/containers/form_group.rb)'s constructor appropriately.

We _could_ also go a stage further and take a hash (like `label:` or `legend:`).

This is the first time the feature has been requested, and there are definitely [workarounds](https://github.com/DFE-Digital/govuk_design_system_formbuilder/issues/153#issuecomment-652380672) if you just want to style a form group in a particular way.

peteryates:
Allow extra classes to be set on form groups via a `form_group_classes` keyword arg on all helpers that wrap contents in a `govuk-form-group`.

Also standardise the other helpers that allow custom `classes` so they all take either an array or a string.

Fixes #158

### DIFF
--- a/spec/support/shared/shared_form_group_examples.rb
+++ b/spec/support/shared/shared_form_group_examples.rb
@@ -1,7 +1,11 @@
 shared_examples 'a field that contains a customisable form group' do
   example_group 'when custom classes are provided' do
     let(:default_class) { %w(govuk-form-group) }
-    subject { builder.send(*args, form_group_classes: custom_classes) }
+    let(:block_content) { %(Are you acquainted with our state's stringent usury laws?) }
+
+    subject do
+      builder.send(*args, form_group_classes: custom_classes) { builder.tag.span(block_content) }
+    end
 
     context 'classes passed in as an array' do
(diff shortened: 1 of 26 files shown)


spec/support/shared/shared_form_group_examples.rb
```diff
@@ -0,0 +1,22 @@
+shared_examples 'a field that contains a customisable form group' do
+  example_group 'when cusom classes are provided' do
```
cpjmcquillan, line 2: `'when cus**t**om classes are provided'`

peteryates: D'oh. Thanks, fixed

PR #161, opened 2020-07-07
Add prefix and suffix text support

### LINKED ISSUE #159: Add prefix and suffix support
**This functionality hasn't been finalised yet so nothing will be merged until the next version of the design system has been released**

It looks like [prefixes and suffixes](https://github.com/alphagov/govuk-frontend/pull/1816) are landing in the next version of the design system.

The approach wraps the input in a `div.govuk-input__wrapper` and places a `span` containing the suffix/prefix accordingly:

```html
  <label class="govuk-label" for="input-with-prefix-attribute">
    Amount
  </label>
  <div class="govuk-input__wrapper">
    <span class="govuk-input__prefix" aria-hidden="true">£</span>
    <input class="govuk-input" id="input-with-prefix-attribute" name="amount" type="number">
  </div>
</div>
```

## Suggested approach

The call should probably look something like this:

```
= f.govuk_number_field(:price_per_kg, label: { text: 'Price per kilogram' }, prefix_text: '£', suffix_text: 'per kg')
```

* [x] add a `prefix_text` and `suffix_text` keyword argument to `Traits::Input`
* [x] enhance `Traits::Input#input` so that
  * [x] it conditionally wraps the input if a group is present
  * [x] it injects the prefix or suffix before or after the `input` field
* [x] add some tests
* [x] update the documentation
* [x] update the guide

## Topics for discussion

Do we need to allow for custom attributes on the prefix and suffix in addition to the input? My gut feeling is that it's over-complicating matters for the most part. This approach can easily be extended to take `prefix` and `suffix` hashes if we later decide to go that route.

## Related

* [GOV.UK Design System docs PR](https://github.com/alphagov/govuk-design-system/pull/1269/)
* [Prefixes and suffixes documentation preview](https://deploy-preview-1269--govuk-design-system-preview.netlify.app/components/text-input/#prefixes-and-suffixes)

peteryates:
This change adds support for the new prefix and suffix text support in GOV.UK frontend. It allows text to be added before and after text inputs in a grey box to give users more context around the data, for example stating the currency of a expected monetary value.

* [x] Wait for version ~`3.8.0`~ `3.9.0` of the Design System to be released and make sure everything looks ok

Fixes #159

### DIFF
--- a/guide/content/form-elements/text-field.slim
+++ b/guide/content/form-elements/text-field.slim
@@ -129,6 +129,16 @@ section
         more details on the reasoning behind this decision in #{link_to("the accompanying blog post", input_types_for_numbers_link).html_safe}.
 
   == render('/partials/example-fig.*',
+    caption: "Inputs with prefixes and suffixes",
+    code: text_field_with_prefix_and_suffix,
+    hide_html_output: false) do
+
+    p.govuk-body
+      | Prefixes and suffixes are useful when you use symbols and abbreviations
+        that are commonly known and understood. Symbols and abbreviations should
+        be explained in the label or hint.
+
+  == render('/partials/example-fig.*',
     caption: "Generating text inputs with custom widths",
     code: text_field_with_custom_width,
     hide_html_output: true) do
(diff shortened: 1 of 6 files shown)


peteryates:
Also worth noting that the docs currently look like this and will do until the guide can be built with the unreleased prefix/suffix styles.

Screenshot from 2020-07-09 09-39-35

peteryates:
Screenshot from 2020-09-14 20-01-23

Looks good!

PR #166, opened 2020-07-29
Investigate adding fixture-based specs

peteryates:
This is an initial attempt at using [GOV.UK Frontend's recent fixtures addition](https://github.com/alphagov/govuk-frontend/issues/1830) to ensure the HTML we're generating is in accordance with the Design System.

After flipping a few default values it's not a million miles off, but there are some snags:

1. Rails insists on [adding a `data-disable-with` attribute to the submit](https://api.rubyonrails.org/classes/ActionView/Helpers/FormTagHelper.html#method-i-submit_tag), even when I attempt to [disable it by passing `false`](https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/166/files#diff-2c10b5e6f11bb65361b9cc8fbca59fcfR47)

2. Because in all cases (other than `#govuk_date_field`) we are merely wrapping the built-in Rails form helpers, we're not in control of the order of attributes. Comparing them directly will fail.

Both of these problems are demonstrated in the RSpec output:

```diff
- <input value="Start now" type="submit" name="start-now" class="govuk-button" data-module="govuk-button">
+ <input type="submit" name="start-now" value="Start now" class="govuk-button" data-module="govuk-button" data-disable-with="Start now" />
```

While using the fixtures would be nice, it's clearly aimed at ports rather than supporting libraries. Getting it all in sync is probably going to be a headache compared to the succinct specs we already have.

Is there a better way of approaching this?

### DIFF
--- a/guide/package-lock.json
+++ b/guide/package-lock.json
@@ -5,9 +5,8 @@
   "requires": true,
   "dependencies": {
     "govuk-frontend": {
-      "version": "3.7.0",
-      "resolved": "https://registry.npmjs.org/govuk-frontend/-/govuk-frontend-3.7.0.tgz",
-      "integrity": "sha512-G3bqoKGGF8YQ18UJH9tTARrwB8i7iPwN1xc8RXwWyx91q0p/Xl10uNywZLkzGWcJDzEz1vwmBTTL3SLDU/KxNg=="
+      "version": "github:alphagov/govuk-frontend#634eeca76ee6f4399cba54cc13a11bd38f5c450a",
+      "from": "github:alphagov/govuk-frontend#634eeca76ee6f4399cba54cc13a11bd38f5c450a"
     }
   }
 }
--- a/guide/package.json
+++ b/guide/package.json
@@ -12,6 +12,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "govuk-frontend": "^3.7.0"
+    "govuk-frontend": "github:alphagov/govuk-frontend#634eeca76ee6f4399cba54cc13a11bd38f5c450a"
   }
 }
--- a/lib/govuk_design_system_formbuilder/builder.rb
+++ b/lib/govuk_design_system_formbuilder/builder.rb
@@ -716,7 +716,7 @@ module GOVUKDesignSystemFormBuilder
     #   = f.govuk_submit "Proceed", prevent_double_click: true do
     #     = link_to 'Cancel', some_other_path, class: 'govuk-button__secondary'
     #
-    def govuk_submit(text = config.default_submit_button_text, warning: false, secondary: false, classes: nil, prevent_double_click: true, validate: false, disabled: false, &block)
+    def govuk_submit(text = config.default_submit_button_text, warning: false, secondary: false, classes: nil, prevent_double_click: false, validate: true, disabled: false, &block)
       Elements::Submit.new(self, text, warning: warning, secondary: secondary, classes: classes, prevent_double_click: prevent_double_click, validate: validate, disabled: disabled, &block).html
     end
 
--- a/lib/govuk_design_system_formbuilder/elements/submit.rb
+++ b/lib/govuk_design_system_formbuilder/elements/submit.rb
@@ -39,9 +39,16 @@ module GOVUKDesignSystemFormBuilder
         {
           formnovalidate: !@validate,
           disabled: @disabled,
+          name: @text.parameterize,
           data: {
             module: %(#{brand}-button),
-            'prevent-double-click': @prevent_double_click
+            'prevent-double-click': @prevent_double_click,
+
+            # this is an attempt to prevent Rails from adding a
+            # data-disable-with attr to the input element.
+            #
+            # it doesn't work (╯°□°）╯︵ ┻━┻
+            disable_with: false
           }.select { |_k, v| v.present? }
         }
       end
--- a/spec/fixtures/button_spec.rb
+++ b/spec/fixtures/button_spec.rb
@@ -1,0 +1,32 @@
+describe GOVUKDesignSystemFormBuilder::FormBuilder do
+  include_context 'setup builder'
+
+  describe 'Button' do
+    # we only generate `<input type=submit...` so ignore everything else like `<button>`
+    applicable = ['input']
+    
+    fixtures = JSON
+      .parse(File.read('guide/node_modules/govuk-frontend/govuk/components/button/fixtures.json'))
+      .select { |f| f['name'].in?(applicable) }
+
+    fixtures.each do |fixture|
+      describe fixture['name'] do
+        let(:expected) { fixture['html'] }
+
+        let(:button_text) { fixture.dig('options', 'text') }
+        subject { builder.govuk_submit(button_text) }
+
+        # name: input
+        # options:
+        #   element: input
+        #   name: start-now
+        #   text: Start now
+        #   html: <input value="Start now" type="submit" name="start-now" class="govuk-button" data-module="govuk-button">
+
+        specify 'should match the HTML' do
+          expect(subject).to eql(expected)
+        end
+      end
+    end
+  end
+end


frankieroberto:
@peteryates it looks like your `govuk_submit` will always out a `<input type="submit">` element, whereas the `govuk_button` nunjucks macro will often output a `<button>` element or a `<a>` element, depending on whether the `element` attribute is set or a `href` is present?

To make use of the fixture files, I'd suggest aligning your form builder more closely with the nunjucks macros, and implementing a `govukButton` method instead (feel free to [borrow ours](https://github.com/UKGovernmentBEIS/govuk-design-system-rails/blob/master/app/helpers/govuk_design_system/button_helper.rb)), and then you could have your `govuk_submit` method wrap that (and perhaps also set the `element` to `input`)?

peteryates:
Thanks for taking a look @frankieroberto. In the spec I'm only pulling out the fixtures that are _applicable_ ([here](https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/166/files#diff-45320b32d0062f8cc0c5d700c366e02aR6)). I tested out just with the basic `input` one but `input attributes`, `input classes` and `input disabled` should be in there too.

As this library is _just_ a form builder I'm more inclined to keep it closely resembling Rails' default than the Nunjucks macros.

This PR was more of an experiment in seeing whether comparing against the fixtures would be of use. I think that comparing two chunks of HTML is likely to be brittle; failing because attributes are in a different order or spacing/indentation don't match is very likely.

For ports it totally makes sense, but this is a bit more of a specialised case imo.

frankieroberto:
@peteryates yeah, it depends on what your formbuilder is trying to achieve, I guess. The approach I've started to take is to have helpers which match the nunjucks macros `govukInput` and `govukButton` and so on, right down to the camelCased method and attribute names. And then to have a form builder class which adds those methods to the form, but binds them to the model (so that the `value` gets set, and the error message are added from the model, and so on).  

This deviates from the standard Rails form helpers, eg in that the labels and the inputs are combined:

```ruby
<%= form_with model: @thing do |form| %>
  <%= form.govuk_input :attribute_name, label: "Blah" %>
<% end %>
```

Depends whether you think it's more useful to follow the Rails conventions or to more closely follow the Design System’s nunjucks API...! 😄 

For what it's worth, I’ve never found Rails’s `form.submit` to be super useful, as there’s rarely a need to bind the submit button to the form, so `tag.button "Continue"` does the same thing, and outputting a `<button>` has some advantages over `<input type="submit">` (mainly in that you can embed HTML within it if needed).

peteryates:
> @peteryates yeah, it depends on what your formbuilder is trying to achieve, I guess. The approach I've started to take is to have helpers which match the nunjucks macros govukInput and govukButton and so on, right down to the camelCased method and attribute names. And then to have a form builder class which adds those methods to the form, but binds them to the model (so that the value gets set, and the error message are added from the model, and so on).

That's a really interesting approach, there's definitely benefits to matching the Nunjucks macros. Does you automatically take care of hyperlinking between items in the error summary and form elements (especially the first radio/checkbox)? That was a _bit of a pain_ for me early on.

> This deviates from the standard Rails form helpers, eg in that the labels and the inputs are combined:
> 
> ```erb
> <%= form_with model: @thing do |form| %>
>   <%= form.govuk_input :attribute_name, label: "Blah" %>
> <% end %>
> ```

Nearly a perfect match for ours:

```erb
<%= form_with model: @thing do |form| %>
  <%= form.govuk_text_field :attribute_name, label: { text: "Blah" } %>
<% end %>
```

I also went down the route of letting the [input, hint, label, caption and legends](https://govuk-form-builder.netlify.app/introduction/labels-captions-hints-and-legends/) be configured via args on the helper. It felt like a cleaner, simpler solution at the time and (thankfully) the design system is quite rigid in that regard!

frankieroberto:
@peteryates re:

> That's a really interesting approach, there's definitely benefits to matching the Nunjucks macros. Do you automatically take care of hyperlinking between items in the error summary and form elements (especially the first radio/checkbox)? That was a bit of a pain for me early on.

I’m taking the approach of making the first item within a checkbox or radio field have the id of the attribute name, with all the subsequent ones having a suffix containing the index number (the same [as the Nunjucks macro does by default](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/radios/template.njk#L55-L60)), and then the error summary just always links to an ID with the attribute name.

The harder thing to do is to make the date component link to the month or year inputs if the error is just with one of those. For now, it always links to the date input (and all three get the red border).

Your label arguments look good, and very similar. I wasn't sure with our implementation whether to always make the label be an object, or to have separate arguments for the other optional values. The advantage of keeping them separate is that it's easier to mark the label as required (as it'll otherwise raise a missing argument error), and it doesn't need to have an `html` or `text` key, which aren't really needed for Rails as it already has its own way of marking strings as html-safe or not. The disadvantage though is that then all the other label arguments need a prefix, ie:

```ruby
<%= form_with model: @thing do |form| %>
  <%= form.govuk_input :attribute_name, 
     label: "Blah", label_classes: "govuk-label--m",
     hint: "Blah", hint_classes: "app-hint--blah" 
  %>
<% end %>
```

Regardless, I think one possible approach would be for a shared gem which _just_ implements the Design System macros as helpers, using the exact same API (eg camelCasing and hashes as arguments) and the fixtures for tests, and then having a separate gem (or gems) which provide a form builder that uses those helpers and binds them to a model, using more Ruby-ish conventions for method names and keyword arguments.

peteryates:
> I’m taking the approach of making the first item within a checkbox or radio field have the id of the attribute name, with all the subsequent ones having a suffix containing the index number (the same as the Nunjucks macro does by default), and then the error summary just always links to an ID with the attribute name.

Ah, makes sense. I took a slightly different approach and automatically insert `error` into the ID when the field has an error. That might be something that I could simplify using the approach you mentioned. I'll take a look and see if there's room for improvement :+1: 

> The harder thing to do is to make the date component link to the month or year inputs if the error is just with one of those. For now, it always links to the date input (and all three get the red border).

Yep, me too. You'd need a new date class to do this, with custom validations that decide _how_ the date is incorrect; and that would depend on context and the rules of the app. It's an entire project's worth of work.

> The advantage of keeping them separate is that it's easier to mark the label as required (as it'll otherwise raise a missing argument error), and it doesn't need to have an html or text key, which aren't really needed for Rails as it already has its own way of marking strings as html-safe or not. The disadvantage though is that then all the other label arguments need a prefix, ie:

I think this is the main reason when I started this gem I took inspiration from the MoJ's elements form builder rather than the Nunjucks macros. It's not intended to be flexible enough for _all_ situations; it's aimed at covering the functionality defined by the design system with an API easy enough to remember (or guess), with as little additional plumbing as possible. The goal is to be sufficient for _most_ fields in _most_ forms, _most_ of the time. [Apply](https://github.com/DFE-Digital/apply-for-teacher-training), a big form-heavy service, use it throughout.

It's also aimed at making the most of Ruby - none of the helpers take `_html` options and I'd strongly suggest that people shouldn't embed HTML in that fashion. Instead, [we use procs](https://govuk-form-builder.netlify.app/introduction/labels-captions-hints-and-legends/#setting-label-content-with-a-proc) instead. The linked example _does_ include inline HTML within the proc as the guide is generated by a static site generator using the form builder itself, but this would work in Rails:

```slim
= f.govuk_text_field :favourite_shade_of_orange,
  label: -> do
    h2.govuk-header-m.orange-underline
      | What’s your favourite shade of orange?
  end
```
Of course within that proc you could call a helper, render a partial or ViewComponent, etc. Easy to write, easy to read, easy to test, syntax highlighting just works etc. All of the things embedded strings of HTML make a pain.

You _could_ build something into your translation layer to convert it into a Nunjucks-compatible JSON document but there'll be a lot of complexity and moving parts. This library is quite small, modular and seemingly easy to grasp.

peteryates:
Closing, may revisit for some components once the final spec/functionality lands.

PR #178, opened 2020-08-14
Fix duplicating content bug

peteryates:
This fixes a bug in `govuk_radio_button_fieldset` and `govuk_checkbox_fieldset` where the block content was rendered twice. It was caused by the removal of `capture` in 08fe4ffc05 and not picked up by the test suite, probably due to it not using a templating engine and using Ruby directly.

There's probably a case for more investigation on this and perhaps some tests that do fire up Slim and ensure things are as they seem.


peteryates:
Thankfully I think most projects' tests would have caught this if it'd been encountered as the radio and checkbox fieldsets would've made Capybara complain that there are duplicate ids. That's how I found it :disappointed:

Issue #177, opened 2020-08-14
Unable to set label content


paulrobertlloyd:
I’m trying to implement an address pattern using multiple text inputs, [as described in the design system](https://design-system.service.gov.uk/patterns/addresses/). This requires the following HTML:

```html
<div class="govuk-form-group">
  <label class="govuk-label" for="address-line-1">
    Building and street <span class="govuk-visually-hidden">line 1 of 2</span>
  </label>
  <input class="govuk-input" id="address-line-1" name="address-line-1" type="text" autocomplete="address-line1">
</div>
```

Looking through [the docs](https://govuk-form-builder.netlify.app/introduction/labels-captions-hints-and-legends/#setting-label-content-with-a-proc), I thought I could achieve this with the following:

```erb
<%= f.govuk_text_field :address_line1, autocomplete: 'address-line1', label: -> do %>
  Building and street <span class="govuk-visually-hidden">line 1 of 2</span>
<% end %>
```

However it appears that this replaces all the content of this helper, both the `label` and the text `input`, whereas the documentation suggests this should only replace the content of the `label` element. Am I misinterpreting the documentation, or is this a bug? /cc @tvararu

tvararu:
Tested this locally, code:

![Screenshot 2020-08-14 at 12 21 46](https://user-images.githubusercontent.com/1650875/90244468-bed66080-de28-11ea-94ee-36503cbef2a0.png)

Result:

![Screenshot 2020-08-14 at 12 21 52](https://user-images.githubusercontent.com/1650875/90244481-c4cc4180-de28-11ea-9bd7-93e44cb61a6e.png)

peteryates:
This happens because the `label` is expecting a [`Proc`](https://ruby-doc.org/core-2.7.1/Proc.html) instead of a block. We can't pass the label a block because `govuk_text_field` already accepts them for [supplementary content](https://govuk-form-builder.netlify.app/building-blocks/injecting-content/) and it would get extremely confusing.

The proc is expecting Ruby rather than ERB, but you can render whatever you like via `render` or calling a helper:

```slim
/_name_label.slim

| Building and street

span.govuk-visually-hidden
  | line 1 of 2
```

```slim
/_form.slim

= form_for @person do |f|

  = f.govuk_text_field :name, label: ->{ render partial: 'name_label' }
```

Alternatively, using a helper:

```ruby
module PeopleHelper
  def name_label
[...]

PR #181, opened 2020-08-15
Test out the generation of ERB input examples

peteryates:
This was brought up in #177 by @paulrobertlloyd where it was mentioned that [Slim](http://slim-lang.com/) templates might not make sense to people who are more used to [ERB](https://ruby-doc.org/stdlib-2.7.1/libdoc/erb/rdoc/ERB.html). In other documentation often multiple languages/syntaxes are listed, this is a quick test to see if it's feasible.

This approach, using Slim's built-in `Slim::ERBConverter` _mostly_ works:

| Slim | ERB |
| ---- | ---- |
| Screenshot from 2020-08-15 11-22-33 | Screenshot from 2020-08-15 11-07-50 |

There are a couple of areas that don't:

* **Blocks** - the `Slim::ERBConverter` conversion is outputting  unwanted text like `_slim_controls1 =`. There might be a way of suppressing it. More investigation needed.

Screenshot from 2020-08-15 11-08-14

* **General formatting**, no space is left before heading tags and after ERB blocks. Using `beatify` half fixes this but causes problems elsewhere, including splitting procs made with arrow syntax across multiple lines. Perhaps there's a better formatting utility out there :man_shrugging: 

Screenshot from 2020-08-15 11-15-25

### DIFF
--- a/guide/content/css/guide.sass
+++ b/guide/content/css/guide.sass
@@ -87,6 +87,10 @@ figure.full-sample
     pre,.example-output
       @include light-grey-border
 
+    .govuk-tabs
+      pre,.example-output
+        border: none
+
     &:last-child
       margin-bottom: 0rem
 
@@ -99,6 +103,10 @@ pre
   > code.highlight
     font-family: FiraMono, monospace
     word-wrap: normal
+
+  > code.wrap
+    white-space: pre-wrap
+    word-wrap: break-word
 
   &.smaller > code.highlight
     font-size: 85%
--- a/guide/layouts/partials/example-fig.slim
+++ b/guide/layouts/partials/example-fig.slim
@@ -44,10 +44,30 @@ figure.full-sample
           | #{localisation}
 
   section
-    h4.govuk-heading-s.example-subheading.input Input
-    pre.example-input
-      code.highlight.language-slim
-        | #{code}
+
+    .govuk-tabs data-module='govuk-tabs'
+      h4.govuk-heading-s.govuk-tabs--title.example-subheading.input Input
+
+      ul.govuk-tabs__list
+        li.govuk-tabs__list-item.govuk-tabs__list-item--selected
+          a.govuk-tabs__tab href=%(#input-slim-#{anchor_id(caption)})
+            | Slim
+        li.govuk-tabs__list-item
+          a.govuk-tabs__tab href=%(#input-erb-#{anchor_id(caption)})
+            | ERB
+
+      .govuk-tabs__panel id=%(input-slim-#{anchor_id(caption)})
+        div
+          pre.example-input
+            code.highlight.language-slim
+              | #{code}
+
+      .govuk-tabs__panel.govuk-tabs__panel--hidden id=%(input-erb-#{anchor_id(caption)})
+        div
+          pre.example-input.smaller
+            code.highlight.language-erb.wrap
+              = format_erb(code)
+
 
   section
     - display_errors = defined?(show_errors) && show_errors
--- a/guide/lib/helpers.rb
+++ b/guide/lib/helpers.rb
@@ -2,6 +2,8 @@ require 'pry'
 require 'action_view'
 require 'active_model'
 require 'htmlbeautifier'
+require 'slim/erb_converter'
+
 Dir.glob(File.join('./lib', '**', '*.rb')).sort.each { |f| require f }
 
 $LOAD_PATH.unshift(File.expand_path("../../lib", "lib"))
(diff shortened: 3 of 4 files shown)


peteryates:
Having investigated this further it seems that converting ERB to Slim is [better supported](https://github.com/slim-template/html2slim) than Slim to ERB. The obvious thing to do would be to switch out the Slim examples and replace them with ERB ones.

Then we can just use `ERB.new(select_field_with_label_and_hint).results_with_hash(args)` on examples like this:

```ruby
module Examples
  module Select
    def select_field_with_label_and_hint
      <<~SNIPPET
        <%= f.govuk_collection_select :new_department_id,
          departments,
          :id,
          :name,
          label: { text: "Which department do you work in?" },
          hint: { text: "You can find it on your ID badge" } %>
      SNIPPET
    end
  end
end
```

Unfortunately, this isn't the case. Rails doesn't actually use regular old [ERB](https://ruby-doc.org/stdlib-2.7.1/libdoc/erb/rdoc/ERB.html). As per [this blogpost](http://timelessrepo.com/block-helpers-in-rails3#impossible_but_it_works_in_rails_3) from the Rails 3 era, Rails _extends_ the [Erubi](https://github.com/jeremyevans/erubi) library in a way that makes expression-printing tags (`<%= 'foo' %>`) with blocks valid. It [appears that it still does some _magic_](https://github.com/rails/rails/blob/master/actionview/lib/action_view/template/handlers/erb.rb) to accomplish this.

As a result, trying to solve the problem with Ruby's built-in ERB is a no-go.

The final option is to attempt to do the same thing but via [Tilt](https://github.com/rtomayko/tilt).

peteryates:
```ruby
Tilt::ErubiTemplate.new { select_field_with_label_and_hint }.render
```

```
[4] pry(#<Nanoc::Core::Context>)> tmpl.render
SyntaxError: (__TEMPLATE__):1: syntax error, unexpected ')'
...t: 'Where do you live?' }) do ).to_s; _buf << '
...                              ^
(__TEMPLATE__):8: syntax error, unexpected `ensure', expecting ')'
; ensure
  ^~~~~~
(__TEMPLATE__):10: syntax error, unexpected `end', expecting ')'
```

Without Rails' modifications it looks like we're failing in the same place.

For clarity, this parses but won't capture the fieldset:

```erb
<% f.govuk_fieldset(legend: { text: 'Where do you live?' }) do %>
  <%= f.govuk_text_field :address_one,   width: 'one-half',    label: { text: 'House number and street' } %>
  <%= f.govuk_text_field :address_two,   width: 'one-half',    label: { text: 'Town', hidden: true } %>
  <%= f.govuk_text_field :address_three, width: 'one-half',    label: { text: 'City' } %>
  <%= f.govuk_text_field :postcode,      width: 'one-quarter', label: { text: 'Postcode' } %>
<% end %>
```
And this fails to parse with the error above but works just fine in Rails:

```erb
<%= f.govuk_fieldset(legend: { text: 'Where do you live?' }) do %>
  <%= f.govuk_text_field :address_one,   width: 'one-half',    label: { text: 'House number and street' } %>
  <%= f.govuk_text_field :address_two,   width: 'one-half',    label: { text: 'Town', hidden: true } %>
  <%= f.govuk_text_field :address_three, width: 'one-half',    label: { text: 'City' } %>
  <%= f.govuk_text_field :postcode,      width: 'one-quarter', label: { text: 'Postcode' } %>
<% end %>
```

Using Rails' implementation in pure Ruby (well, Nanoc) appears to be quite an involved process. It appears that the original approach of converting Slim to ERB might be the best option :unamused:

peteryates:
I've made progress on this, ERB examples are actually looking pretty decent. Not _perfect_ but definitely usable.

| Slim | Erb |
| ----- | ----- |
| Screenshot from 2020-08-30 19-41-06 | Screenshot from 2020-08-30 19-41-13 |

paulrobertlloyd:
🎉

peteryates:
This is going to be part of version 2 so the ERB-ified documentation can be viewed [here](https://deploy-preview-175--govuk-form-builder.netlify.app/).

Please feel free to take [the beta](https://rubygems.org/gems/govuk_design_system_formbuilder/versions/2.0.0b2) for a spin, I have [a fork of Apply](https://github.com/graphia/apply-for-teacher-training/tree/upgrade-to-govuk-design-system-formbuilder-v2.0.0) that I'm testing against and with a bit of finding and replacing (mainly `hint_text: 'blah'` ➡ `hint: { text: 'blah' }`) everything appears to work ok.

PR #191, opened 2020-09-03
Fix localisation of collection labels/hints

peteryates:
This fixes a bug where @value wasn't being assigned when the hint_method param wasn't passed in, even though @value is required to localise it properly.

Tests have been added to ensure it works as expected; for radio collections both the label and hint can be set via localisation and for check boxes just the hint can.

Thanks @zheileman for reporting and diagnosing :+1:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,6 +9,8 @@ AllCops:
     - govuk_design_system_formbuilder.gemspec
 Rails/Date:
   Enabled: false
+Rails/ActiveSupportAliases:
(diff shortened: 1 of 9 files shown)


zheileman:
This is great. Thank for that!

peteryates:
@zheileman hopefully your suite willl pass with [2.0.0b3](https://rubygems.org/gems/govuk_design_system_formbuilder/versions/2.0.0b3 ) :crossed_fingers:

PR #200, opened 2020-09-29
Add missing legend to guide proc example

peteryates:
In this case the proc completely overwrites the legend so the wrapping tag needs to be provided manually.


fofr (approved):

fofr:
I wonder if it's worth noting the `legend` tag must be provided in the guidance text too.

peteryates:
Screenshot from 2020-09-29 13-22-20

I _think_ this covers it.

Issue #201, opened 2020-09-29
Legends should not default to using an H1


fofr:
Including an H1 in a legend by default is not semantically correct.

When writing:

```ruby
legend: { text: "Legend text" }
```

The following HTML is output, defaulting to using an H1 tag:
```html
<legend class="govuk-fieldset__legend govuk-fieldset__legend--m">
  <h1 class="govuk-fieldset__heading">Legend text</h1>
</legend>
```

When using the official Nunjucks macro:

```js
fieldset: {
  legend: {
    text: "Legend text"
  }
}
```

The default HTML does not include an H1:

```html
<legend class="govuk-fieldset__legend">
  Legend text
</legend>
```

Headings only get included with the `isPageHeading` flag:

https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/fieldset/template.njk#L8-L14

edwardhorsford:
Strong agree - ideally we'd have the same defaults as the design system.

aliuk2012:
I agree, it should be an optional setting called something like "isPageHeading" - [source](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/fieldset/fieldset.yaml#L23-L26)

peteryates:
Yeah, this makes sense. I originally built it based purely off the examples on [this page](https://design-system.service.gov.uk/components/fieldset/), both of which wrap it in `<h1>`. As I never really looked into the Nunjucks it didn't occur to me to allow non-header legends.

It's possible to set the tag to another header level by passing in `tag: 'h3'` or whatever in the legend options, and also possible to set the default tag in the config:

```ruby
GOVUKDesignSystemFormBuilder.configure do |conf|
  conf.default_legend_tag  = 'h3'
  conf.default_legend_size = 'l'
end
```

Unfortunately, it's currently not possible to omit the `tag` altogether, so I think once that's implemented all of the desired functionality will be present. The equivalent of `isPageHeading` will be achievable by passing in `tag: 'h1'` and the default behaviour will be no tag (`tag: nil`) which will just output `<legend class="govuk-fieldset__legend">Legend text</legend>`.

fofr:
@peteryates Agree about the examples being misleading, I've raised this issue: https://github.com/alphagov/govuk-design-system/issues/1364

peteryates:
Fixed by #204, will be included in `v2.1.0` shortly.

PR #204, opened 2020-09-29
Remove default header tag from legends

peteryates:
As per #201, legends shouldn't contain `<h1>` tags by default.

This change modifies the behaviour of legends so that when `tag: nil` the legend content isn't wrapped in any other tags. To override this behaviour, ie when the heading is also the page title, tag can be set to `tag: 'h1'`. to achieve the old behaviour.

The new value for `default_legend_tag` is `nil`.

This is somewhat related to https://github.com/alphagov/govuk-design-system/issues/1364

### Example output:

```slim
= f.govuk_date_field :date_of_birth,
  date_of_birth: true,
  legend: { text: 'Enter your date of birth' },
  hint: { text: 'For example, 31 3 1980' }
```

```html
  <fieldset class="govuk-fieldset" aria-describedby="person-date-of-birth-hint">
    <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">
      Enter your date of birth
    </legend>
    <span class="govuk-hint" id="person-date-of-birth-hint">
      For example, 31 3 1980
    </span>
...
```

### DIFF
--- a/guide/content/introduction/configuration.slim
+++ b/guide/content/introduction/configuration.slim
@@ -21,7 +21,7 @@ p.govuk-body
 
           conf.default_submit_button_text   = 'Apply'                       # Continue
           conf.default_error_summary_title  = 'Uh-oh, spaghettios'          # There is a problem
-          conf.default_legend_tag           = 'h3'                          # h1
+          conf.default_legend_tag           = 'h3'                          # nil
           conf.default_legend_size          = 'l'                           # m
           conf.default_caption_size         = 'xl'                          # m
           conf.default_radio_divider_text   = 'how about'                   # or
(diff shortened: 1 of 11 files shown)


fofr:
Example output looks good. Thanks for the super fast turnaround!

Code looks good to me, but you might want someone with more of an eye for Ruby to check it too.

peteryates:
Thanks for checking. The _actual_ code change is only about five lines and is very straightforward so I'm confident about it; the rest is tests (and typo fixes).

Also, thank you for raising this. It's the kind of bug that I just wasn't aware of and it'll benefit everyone using the library. Please continue to report anything else that looks iffy 😅

Issue #203, opened 2020-09-29
govuk_submit shoudl output a button not an input type=submit


aliuk2012:
@edwardhorsford raised a good point here https://github.com/DFE-Digital/register-trainee-teacher-data/pull/58#discussion_r496082292 that `form.govuk_submit` method should output a `<button>` instead of `<input type="submit"...>`. 

There is a benefit of using the button as you can also use html inside the button which could be set using a block. by default a button inside a form should submit the form by default

peteryates:
As discussed in #207 this is probably better served with a component or directly in the template.

PR #205, opened 2020-09-30
Take a quick look at button tag generation

peteryates:
I'm in two minds about whether or not the presence of a `button` param is a good idea. Unless there's HTML inside the button, is there any advantage of a `<button>` over an `<input type='submit'>`?

Leaning towards choosing the tag based on content (which avoids the problem of dealing with passing in a proc when `button: false`) and having a `config.submit_generate_buttons = (false|true)` to globally override for the sake of developer preference.

Refs #203

### DIFF
--- a/lib/govuk_design_system_formbuilder/builder.rb
+++ b/lib/govuk_design_system_formbuilder/builder.rb
@@ -799,7 +799,8 @@ module GOVUKDesignSystemFormBuilder
 
     # Generates a submit button, green by default
     #
-    # @param text [String] the button text
+    # @param content [String,Proc] the submit input or button text. When a Proc is provided the contents will be rendered inside a +<button>+ tag instead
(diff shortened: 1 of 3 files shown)


peteryates:
For reference, there are actually some differences between `<button>` and `<input type='submit'>` in IE6, [namely that](http://vancelucas.com/blog/ie6-and-multiple-button-submit-elements/):

> IE6 actually passes ALL `<button type="submit">` elements back to the server on postback instead of just the active/clicked one. As a result, the “update” element will ALWAYS be set for ie6 users, and the “else” condition will NEVER be executed. They won’t be able to make it past the cart page, and won’t be able to buy your products or continue to the next step.

I know IE6 is no longer supported, am quite happy to change out this behaviour if deemed necessary.

peteryates:
Replaced by #207

PR #207, opened 2020-10-01
Render button when HTML content provided to submit

### LINKED ISSUE #203: govuk_submit shoudl output a button not an input type=submit

peteryates:
The `#govuk_submit` method usually renders an `<input type='submit'>`.

Sometimes it might be useful to allow HTML to be included inside a form submission button which would be supported by a `<button>` tag instead.

When the submit content is just text it seems that there's no advantage in using a button over an input, so we'll stick to using what Rails expects.

This change updates the behaviour so when `#govuk_submit` is supplied with HTML (via a proc), it render the contents inside a button tag.

This deviates slightly from Rails' form builder which always uses an input, and there *may* be issues with the way IE6 deals with buttons, but it shouldn't cause any major problems with modern browsers.

It's worth noting that the main reason for opting for a proc instead of a block here is that blocks are already used by `#govuk_submit` to position subsequent buttons and apply spacing between them.

### With a string

```slim
= f.govuk_submit 'Do the thing'
```
This will render

```html
<input class="govuk-submit" value="Do the thing"/>
```

### With a proc

```slim
= f.govuk_submit -> { tag.strong('Do the thing') }
```

This will render.

```html
<button class="govuk-submit">
  <strong>Do the thing</strong>
</button>
```

Buttons placed inside forms will automatically submit them.

---------

Fixes #203
Refs #205

### DIFF
--- a/guide/content/css/guide.sass
+++ b/guide/content/css/guide.sass
@@ -185,6 +185,15 @@ p, li, dt, dd, td, th
   padding: 0 0.2rem
   background: linear-gradient(90deg, rgba(184,255,255,1) 0%, rgba(255,152,255,1) 50%, rgba(255,255,108,1) 100%)
 
+.example-output > button
+  padding: 0.2em 1em 1em
+  font-size: 150%
+
+  div.large
+    font-weight: bold
+    padding: 0.5em 0.6em 0.2em
+    font-size: 200%
+
 .orange-underline
   text-decoration: underline orange solid 2px
   text-underline-offset: 4px
--- a/guide/content/form-elements/submit.slim
+++ b/guide/content/form-elements/submit.slim
@@ -48,4 +48,14 @@ section
         passed in as a block will be rendered in-line with the primary submit
         button.
 
+  == render('/partials/example-fig.*',
+    caption: "A submit button containing HTML",
+    code: button_containing_html) do
+
+    p.govuk-body
+      | To add HTML content to a button pass in a proc instead of a string.
+        This will render the proc content wrapped in <code>button</code> tags
+        instead of the usual <code>input</code> and offer some additional styling
+        options.
+
 == render('/partials/related-info.*', links: submit_info)
--- a/guide/lib/examples/submit.rb
+++ b/guide/lib/examples/submit.rb
@@ -31,5 +31,13 @@ module Examples
             ' Safe as draft
       SNIPPET
     end
+
+    def button_containing_html
+      <<~SNIPPET
+        = f.govuk_submit -> do
+          div.large Submit
+          ' your application
+      SNIPPET
+    end
   end
 end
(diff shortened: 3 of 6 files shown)


aliuk2012:
I'm not sure about the syntax here. If I just wanted to output a button I would have to remember to use a proc.  Perhaps what is needed is a `govuk_button` component.

- If there was a `govuk-button` component in govuk-components when we wouldn't need to change this gems `govuk_submit` helper.  
- If a service is using both formbuilder and govuk-components they could just use `govuk_button` component. 
- new `govuk_button` could even replace the existing start button component in govuk-component. 
- `govuk_button` could be used to button that might be clickable hooks for javascript by setting the `type` attribute value to `button`

In the govuk-frontend their button component outputs two different tags depending on if an option is set. If href is set then it a link otherwise its a button [source](https://github.com/DFE-Digital/govuk-components/blob/master/app/components/govuk_component/start_now_button.rb). I'd be happy to contribute this component to govuk-components if you think it would be worth it.

peteryates:
I'd be hesitant about replacing the start button component as they tend to actually be links styled at buttons that direct people to the start of a wizard. Changing that for a button in a form that does a `POST` doesn't seem right to me. I'm all for there being a `govuk_button` in the components gem though :+1:

edwardhorsford:
@peteryates in the design system, the single component `govukButton` outputs _either_ a `button` or a `link` depending on if an `href` is passed. So the start button *wouldn't* output a button with a form.

peteryates:
@edwardhorsford - yeah, I was referring to the Rails methods we're wrapping in the components helper. Rails' `#button_to` automatically generates a form containing a `<button>`. Currently the [`GovukComponent::StartNowButton` component](https://github.com/DFE-Digital/govuk-components/blob/master/app/components/govuk_component/start_now_button.html.erb) has no logic and merely wraps `#link_to` and adds on the necessary classes and injects an SVG.

edwardhorsford:
One other thing to note - it should support button links that don't have the arrow. If `isStartButton` is not set to true, the svg shouldn't get injected and you should just get a link with relevant attributes.

peteryates:
So even though this could also be well-served by a component I don't see a problem with including the functionality as it's in-keeping with the uses of procs in other helpers.

Screenshot from 2020-10-02 22-47-22

peteryates:
Having thought about this a bit more, it adds extra complexity to the tests and as it's so easily achieved via a component or right there in the template it's not worth it.

```slim
button.govuk-button
  strong Hello
  | world
```

Issue #208, opened 2020-10-18
Errors on the base object result in broken summary link


FriedSock:
If there is an error that can't be attributed to a specific field, the link for the error will be of the form `#object-base-field-error`, for which there is no element with that id on the page. In this case I think the form should add this id to the first input field in the form, or to the form itself.

peteryates:
Hi, thanks for reporting this. Your suggestion fits in with that of the design system recommendation: 

> If you do not know which field contains an error, link to the first field.

Unfortunately, due to the way the underlying Rails form builder is written, it's _really difficult_ to know which the first field rendered will be when we're rendering the `#govuk_error_summary`.

Linking to the form itself might be more-straightforward; the gem could provide a helper method to return an appropriate `id`.

I'll have a think on it. Suggestions welcome.

peteryates:
@FriedSock - would you mind providing a bit of context around your particular use case? We had a discussion internally and found that in most instances a failing form submission should be tied to at least one of the inputs.

FriedSock:
@peteryates Yes, thank you for looking in to this so quickly.

The case is: a user is required to submit an email address *or* a phone number, when they don't supply either there is a validation error on the base object, I don't think it would be correct to add errors to both of the fields since neither are strictly required by themselves.

peteryates:
@FriedSock - please let me know your thoughts on #209. I feel like this is the best approach.

PR #209, opened 2020-10-23
Allow base errors to link to supplied fields 

peteryates:
In the case where errors are added to the object's `:base`, linking items in the error summary list to a field is impossible. The GOV.UK Design System guidance states that [such errors should link to the first field](https://design-system.service.gov.uk/components/error-summary#linking-from-the-error-summary-to-each-answer).

This change adds a `link_base_errors_to` argument to the error_summary helper that will override the link when errors on the object base are detected.

There's a [new section on the guide](https://deploy-preview-209--govuk-form-builder.netlify.app/introduction/error-handling/#linking-errors-on-the-base-object-to-specific-fields) covering the behaviour in more detail.

Refs #208

### DIFF
--- a/guide/content/introduction/error-handling.slim
+++ b/guide/content/introduction/error-handling.slim
@@ -23,10 +23,25 @@ p.govuk-body
     extra configuration.
 
 == render('/partials/example-fig.*',
-  caption: "A text field with an error",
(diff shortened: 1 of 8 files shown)


FriedSock:
LGTM 👍 

Thank you for doing this.

PR #211, opened 2020-10-27
Propagate value to hint element in FieldsetCheckBox

zheileman:
The value was missing from the `hint_options` hash, and the consequence was (because a `nil` value) the fallback to the hint text of the fieldset, when declaring individual checkboxes inside a `govuk_check_boxes_fieldset` (for example if one of the check boxes has to reveal a text area)

The code causing the issue was (it was entering in the second branching due to value being `nil`):

```ruby
def schema_path
  if @value.present?
    [@object_name, "#{@attribute_name}_options", @value]
  else
    [@object_name, @attribute_name]
  end
end
```

### Example before
Screenshot 2020-10-26 at 10 59 52

### Example after
Screenshot 2020-10-27 at 11 35 26

If any of the individual check boxes had a hint text declared in the locales, it will show as expected:
Screenshot 2020-10-27 at 11 36 15

### DIFF
--- a/guide/content/building-blocks/localisation.slim
+++ b/guide/content/building-blocks/localisation.slim
@@ -58,6 +58,12 @@ p.govuk-body
     sample_data: departments_value_data_raw)
 
   == render('/partials/example-fig.*',
+    caption: "A more comprehensive example of localised check boxes",
+    localisation: movie_genre_check_boxes_locale,
+    code: movie_genre_check_boxes) do
+    == render('/partials/fieldset-warning.*', input_type: 'check box')
+
+  == render('/partials/example-fig.*',
     caption: "Customising locale structure",
     localisation: custom_locale,
     code: role_name,
--- a/guide/lib/examples/localisation.rb
+++ b/guide/lib/examples/localisation.rb
@@ -83,6 +83,42 @@ module Examples
       LOCALE
     end
 
+    def movie_genre_check_boxes
+      <<~SNIPPET
+        = f.govuk_check_boxes_fieldset :movie_genres do
+          = f.govuk_check_box :movie_genres, :action, link_errors: true
+          = f.govuk_check_box :movie_genres, :comedy
+          = f.govuk_check_box :movie_genres, :horror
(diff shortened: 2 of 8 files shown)


peteryates:
Great, thanks for this. I definitely think a spec or two to cover this is a good idea.

zheileman:
@peteryates I tried to add an example to the guide, but for some reason couldn't get it to work this time. I've upgraded my system recently so I think some dependency is now broken, I got this:

```✗ make build-guide
cd guide; npm ci --silent
added 1 packages in 0.637s
cd guide; bundle exec nanoc
Unable to load the EventMachine C extension; To use the pure-ruby reactor, require 'em/pure_ruby'

Captain! We’ve been hit!

LoadError: dlopen(/Users/jesuslaiz/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/eventmachine-1.2.7/lib/rubyeventmachine.bundle, 9): Library not loaded: /usr/local/opt/openssl/lib/libssl.1.0.0.dylib
  Referenced from: /Users/jesuslaiz/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/eventmachine-1.2.7/lib/rubyeventmachine.bundle
  Reason: image not found - /Users/jesuslaiz/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/eventmachine-1.2.7/lib/rubyeventmachine.bundle```

We have smoke tests with markup fixtures that raised this issue when trying to use the latest version of the gem (in one of our projects we were still in the 1.2.x branch) 👍

peteryates:
Ah. your Ruby is probably compiled against the older library. I expect reinstalling Ruby `2.6.5` will fix that for you.

peteryates (approved):
Brilliant, thank you :zombie_man:

cpjmcquillan:
@peteryates I believe there is a similar issue with `govuk_collection_check_boxes` adding the localised hint text into the generated fieldset if it isn't provided rather than respecting something blank

PR #213, opened 2020-10-31
Allow suppression of elements by passing in nil

### LINKED ISSUE #212: Collection check boxes hint reverts to translation file even when hint is nil
When specifying `hint: { text: nil }` as per [the docs](https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder/Builder#govuk_collection_check_boxes-instance_method), the hint text should **not** be rendered, but instead it falls back to the translation `'helpers.hint.{form}.{attribute_name}'`. The current workaround is to ensure the translation isn't present

### Example
image

peteryates:
This addresses a problem where elements are always rendered if there's a matching localisation key, even if `nil` is supplied to the helper.

Introducing a `Elements::Null` class simplifies the logic - an alternative solution was to introduce an `:enabled` flag but as we already have :active it was noisy and confusing.

When the relevant argument (`:caption`, `:hint`, `:label` or `:legend`) is nil a `Elements::Null` is built instead of the usual 'active' element. It will never render any HTML and allows the suppression of elements.

The default argument for captions, hints, labels and legends is always an empty hash, so simply not providing any arguments will allow the localisation mechanism to kick in and populate the element.

Fixes #212

### DIFF
--- a/lib/govuk_design_system_formbuilder/containers/fieldset.rb
+++ b/lib/govuk_design_system_formbuilder/containers/fieldset.rb
@@ -30,7 +30,11 @@ module GOVUKDesignSystemFormBuilder
       end
 
       def legend_element
-        @legend_element ||= Elements::Legend.new(@builder, @object_name, @attribute_name, **legend_options)
+        @legend_element ||= if @legend.nil?
+                              Elements::Null.new
+                            else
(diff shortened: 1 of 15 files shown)


cpjmcquillan (approved):
Looks good to me, thanks 💫

PR #220, opened 2020-11-23
Add support for wildcards in day and month inputs

peteryates:
The [latest version of the design system](https://github.com/alphagov/govuk-frontend/releases/tag/v3.10.0), introduces support for date inputs that [allow an 'X' to be used in place of an exact date](https://github.com/alphagov/govuk-frontend/pull/1975).

This change implements that functionality but does so in a more-restrictive way; users won't have free reign over the pattern but
can optionally have 'X' added to it by supplying `wildcards: true` as an argument.

Should this be more flexible? Is it even necessary? These special 'dates' will need to be intercepted before Rails tries to parse them anyway, perhaps this should be left as an exercise for the developer 🤔

### DIFF
--- a/lib/govuk_design_system_formbuilder/builder.rb
+++ b/lib/govuk_design_system_formbuilder/builder.rb
@@ -850,6 +850,7 @@ module GOVUKDesignSystemFormBuilder
     # @option caption size [String] the size of the caption, can be +xl+, +l+ or +m+. Defaults to +m+
     # @option caption kwargs [Hash] additional arguments are applied as attributes on the caption +span+ element
     # @param omit_day [Boolean] do not render a day input, only capture month and year
+    # @param wildcards [Boolean] add an 'X' to the date wildcards so users can add approximate dates
     # @param form_group [Hash] configures the form group
     # @option form_group classes [Array,String] sets the form group's classes
     # @option form_group kwargs [Hash] additional attributes added to the form group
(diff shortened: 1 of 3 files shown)


zheileman (approved):
I think it looks great. I'm happy with this solution as I'd rather have control and specifically mark a date as "wildcard".

zheileman:
I personally never came across those X dates to be honest, but don't think are very user friendly.

We've implemented some alternatives to this problem where the user might not know the exact date.
Here are some examples.

We have a service where a user can enter an approximate date if they want or if they don't remember the exact date.
We just indicate so in a hint text as well as to reinforce this, we have a check box they can select "This is not the exact date" (we really do nothing different in terms of calculations if they select this check box, but serves us for analytics purposes, to know how many of the dates users enter are exact vs. approximate).

Screenshot 2020-11-24 at 11 13 35

In this service, an X date will not work as we still need to do calculations on dates to estimate the rehabilitation period and date the caution or conviction become spent.
Of course, being an approximate date, will give approximate results, but the user is aware of this and we tell them in the results page as well (the check box is handy for that).

And in another service we use a different approach, in particular for DOB where if they do not know the DOB of a person they can select a check box that reveals a little input to enter either a year, or approx. age:

Screenshot 2020-11-24 at 11 17 55

These are just some examples of a different approach to the problem.

cpjmcquillan (approved):
Seems like a bit of a strange one from the design system, but the way you're supporting it here makes sense to me 👍

PR #227, opened 2020-12-18
Add include_hidden argument to radios collection

peteryates:
Rails has had this functionality [since version 5](https://github.com/rails/rails/pull/18303) so it makes sense for it to be supported here. This change uses the same keyword, `include_hidden`, as Rails so should feel familiar to users. It's intended to [allow forms to be submitted without any of the radios being selected](https://bigbinary.com/blog/rails-5-add-a-hidden-field-on-collection-radio-buttons).


ethax-ross (approved):
👏

zheileman (approved):
💯

PR #229, opened 2020-12-29
Improve Github Actions

peteryates:
This is an attempt to remove some of the manual work required to prep releases and make the output in the actions tab on GitHub a little easier to follow:

* [x] Remove `if` statements from the `tests.yml` workflow
* [ ] Add a new workflow file that can trigger the building and pushing of a gem to RubyGems and GitHub Packages when a tag is specified. This should be a gated (manual) task.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,32 @@
+name: Publish
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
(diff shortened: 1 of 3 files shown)


peteryates:
[Example of it working and pushing to RubyGems](https://github.com/DFE-Digital/govuk_design_system_formbuilder/runs/1621942018?check_suite_focus=true).

PR #233, opened 2021-01-12
Switch to vanilla rubocop

peteryates:
**This is just an experiment - it follows on from #232**

Switch from the GOV.UK rubocop config to the vanilla one as part of the Ruby 3.0.0 upgrade. Seeing as many of the
less-useful cops are already ignored and this has pointed out some improvements to the codebase it feels like a reasonable step.

Leaving here for discussion closer to the time when we're ready to make the jump to supporting 3.0.0.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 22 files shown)


peteryates:
This is too big a leap, let's stick with the GOV.UK variant for the time being.

PR #235, opened 2021-01-15
Fix radio button procs example in guide

peteryates:
There's a bug in the guide where clicking the radio button labels [in the last example](https://govuk-form-builder.netlify.app/form-elements/radios/#populating-values-labels-and-hints-with-procs) doesn't focus the right radio.

It's because the passed-in values are symbols rather than strings.

When a value is passed into the collection helper via a proc as a symbol, it wasn't applied properly when constructing the elements within. This could lead to inputs not having a value, labels not linking to the corresponding input and hints not describing their intended target.

This change ensures that any symbols are converted to strings and adds some tests ensuring that when symbols are supplied via procs the output is rendered properly.

* [x] Check it still works ok when HTML safe strings are passed in

### DIFF
--- a/guide/lib/examples/radios.rb
+++ b/guide/lib/examples/radios.rb
@@ -87,8 +87,8 @@ module Examples
         = f.govuk_collection_radio_buttons :laptop,
           laptops,
           ->(option) { option },
-          ->(option) { I18n.t('laptops.names.' + option.to_s) },
-          ->(option) { I18n.t('laptops.descriptions.' + option.to_s) },
+          ->(option) { I18n.t('laptops.names.' + option) },
+          ->(option) { I18n.t('laptops.descriptions.' + option) },
           legend: { text: "Which laptop would you like to use?" }
       SNIPPET
     end
--- a/guide/lib/setup/example_data.rb
+++ b/guide/lib/setup/example_data.rb
@@ -72,7 +72,7 @@ module Setup
 
     def laptops_data_raw
       <<~DATA
-        laptops = %i(thinkpad xps macbook_pro zenbook)
+        laptops = %w(thinkpad xps macbook_pro zenbook)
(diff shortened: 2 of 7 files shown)


peteryates:
Double checked that this change doesn't affect HTML-safe strings:

This input:

```erb
<%= form.govuk_collection_radio_buttons(
  :department_id,
  %i(technology sales marketing) ,
  -> (item) { item },
  -> (item) { tag.span(item.capitalize, class: %(#{item}-fancy-label)) },
  legend: { text: "Which department do you work in?" }
) %>
```

```html
  <fieldset class="govuk-fieldset">
    <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">Which department do you work in?</legend>
    <div class="govuk-radios" data-module="govuk-radios">
      <div class="govuk-radios__item">
        <input id="person-department-id-technology-field" class="govuk-radios__input" type="radio" value="technology" name="person[department_id]" />
        <label for="person-department-id-technology-field" class="govuk-label govuk-radios__label"><span class="technology-fancy-label">Technology</span></label>
      </div>
      <div class="govuk-radios__item">
        <input id="person-department-id-sales-field" class="govuk-radios__input" type="radio" value="sales" name="person[department_id]" />
        <label for="person-department-id-sales-field" class="govuk-label govuk-radios__label"><span class="sales-fancy-label">Sales</span></label>
      </div>
      <div class="govuk-radios__item">
        <input id="person-department-id-marketing-field" class="govuk-radios__input" type="radio" value="marketing" name="person[department_id]" />
        <label for="person-department-id-marketing-field" class="govuk-label govuk-radios__label"><span class="marketing-fancy-label">Marketing</span></label>
      </div>
    </div>
  </fieldset>
</div>
```

Everything's working as expected 😌

PR #237, opened 2021-02-10
Upgrade to GOV.UK Frontend version 3.11.0

peteryates:
The latest revision to the design system brings one change that affects this library, [button groups](https://github.com/alphagov/govuk-frontend/pull/2114).

Previously the behaviour was to automatically apply a right margin to the submit input when additional content was passed into `govuk_submit` via a block.

The new behaviour is to wrap the submit input and the block contents in a `div.govuk-button-group`; this will allow the buttons and links within to be nicely arranged.

Screenshot from 2021-02-10 15-48-10


zheileman (approved):
💯 🥳 Feels very clean!

PR #241, opened 2021-02-16
Allow novalidate to be enabled by default on forms

peteryates:
This is a rehash of a _really old_ feature (28ffa91071046) whereby the behaviour of `#form_for` and `#form_with` could be altered to default `novalidate: true`. It improves on the previous attempt by wrapping the functionality in conditional logic which requires the `default_form_novalidate` configuration option to be set.

It defaults to `false`.

When it is set to `true` the `formnovalidate` attribute will be omitted from the submit button.


dankmitchell (commented):
Looks fine to me

PR #245, opened 2021-02-25
Add example of hard-coding a radio option hint text

david-mears-dfe:
This alters: https://govuk-form-builder.netlify.app/form-elements/radios/#radio-buttons-fieldset-with-conditional-content-and-a-divider

When adding a text-divider, the list of options has to be hard-coded. It was unclear how to add hint-text to the option in a hard-coded way, since the mentions of hint text for radio options on this page were about doing so programmatically. This change adds an example of how to do this.


peteryates:
Great, thanks for this!

Issue #249, opened 2021-03-15
Idea: custom data attributes on components


ethanmills:
Would you be open to adding the ability to set custom data attributes on components? We're using `data-test` attributes for Cypress tests, and it would be great to be able to set them on formbuilder components. I would be happy to write a PR if that helps.

peteryates:
Yeah, this is definitely something that should be supported consistently across the helpers. Some already support this behaviour by adding any extra keyword args as HTML attributes, but a couple of the mimic Rails' built-in `html_attributes`.

peteryates:
I had a quick play with this and think the best route is to create a `Traits::HTMLAttribute` module that can be dropped in to any of the classes that would support this behaviour, which will probably be most (or all) of the `Elements`.

There's a slight inconsistency because `#govuk_select` mimics Rails' `html_options` arg instead of just using the keyword args for HTML attributes. I guess at this point the key choice is on the direction we take, add a `html_options` arg to the other helpers or allow `#govuk_select` to use the keyword args (and eventually deprecate `html_options`).

ethanmills:
I suspect keyword args will lead to lower maintenance burden in the long run

PR #250, opened 2021-03-16
Pass through additional attributes where not already supported

callumacrae:
Resolves https://github.com/DFE-Digital/govuk_design_system_formbuilder/issues/249 cc @ethanmills 

This adds support for passing through additional attributes where it wasn't already supported, and improves the tests in a couple places where it was already supported but not tested.

Some places already supported basically the same thing but with `html_attributes` instead - I left that alone.

### DIFF
--- a/spec/govuk_design_system_formbuilder/builder/select_spec.rb
+++ b/spec/govuk_design_system_formbuilder/builder/select_spec.rb
@@ -61,7 +61,8 @@ describe GOVUKDesignSystemFormBuilder::FormBuilder do
       let(:extra_args) do
         {
           required: { provided: true, output: 'required' },
-          autofocus: { provided: true, output: 'autofocus' }
+          autofocus: { provided: true, output: 'autofocus' },
+          data: { test: 'abc' }
         }
       end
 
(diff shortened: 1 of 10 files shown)


spec/govuk_design_system_formbuilder/builder/select_spec.rb
```diff
@@ -70,7 +71,13 @@
       specify 'select tag should have the extra attributes' do
         select_tag = parsed_subject.at_css('select')
         extract_args(extra_args, :output).each do |key, val|
-          expect(select_tag[key]).to eql(val)
+          if key == 'data' && val.respond_to?('each')
+            val.each do |dataKey, dataVal|
+              expect(select_tag[%(data-#{key})]).to eql(dataVal)
```
callumacrae, line 76: The linter has pointed out that `dataKey` is unused and I'm using the wrong variable here - why is this test passing?

peteryates: The linter is complaining because in Ruby we'd typically use snake case (e.g., `data_key`).

This bit of code is actually never reached, the value returned from `extract_args(extra_args, :output)` is `{:required=>"required", :autofocus=>"autofocus", :data=>nil}`, so it's skipping to the `else` below.

I'll pick it up from here.

callumacrae: Great thanks!

peteryates:
I'm going to close this as #251 includes the changes made here.

PR #251, opened 2021-03-17
Allow HTML attributes to be set via keyword args across all helpers

### LINKED ISSUE #249: Idea: custom data attributes on components

peteryates:
This is quite a big change but adds plenty of flexibility and standardises the way we assign HTML attributes.

Previously some helpers automatically applied any unrecognised keyword arguments as HTML attributes on the input element, but `#govuk_select` worked differently and they were assigned using `html_options` (like Rails).

Now, all the helpers that output a single input element (ie, everything except the radio button and check box collections and the date field which is comprised of 3 inputs) can have their attributes set using keyword args:

```
= govuk_file_field :photo, accept: ".jpg,.jpeg,.tiff,.heif"
```

This is inkeeping with the behaviour of `label`, `hint`, `legend`, `caption` and `form_group` hashes that already have this functionality.

Fixes #249, #250, #252

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,5 @@
 # .rubocop.yml
+---
 inherit_gem:
   rubocop-govuk:
     - config/default.yml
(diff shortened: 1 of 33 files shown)


cpjmcquillan (approved):
Nice!

PR #255, opened 2021-03-31
Allow date inputs to set default values for multi-parameter attributes that aren't date objects

cpjmcquillan:
In teaching vacancies we use a form model pattern which means sometimes the form builder renders date inputs where the builder object date attribute doesn't respond to `:day`, `:month`, `:year`. 

Rather than default values being blank, it would be nice to fallback to retrieving the values from the multi-parameter hash. 

Another approach would be for us to initialise date objects in our form models, but I'd rather the form builder supported this!


peteryates (approved):
Looks good to me.

PR #256, opened 2021-04-03
Improve HTML attribute merging

peteryates:
Now the formbuilder is using [deep_merge](https://github.com/danielsdeleo/deep_merge) internally we run into a problem with some HTML attributes, namely those that represent lists of values as strings. The most common of these are `class`, `aria-describedby` `aria-labelledby`, but there are probably others too - these all contain a space-separated list of values.

The problem is that when we try to `merge` or `deep_merge` strings, one gets overwritten:

```ruby
{class: "red stripes"}.deep_merge({class: "blue spots"})
=> {:class=>"blue spots"}
```
Instead, if we use arrays:

```ruby
{class: %w(red stripes) }.deeper_merge({ class: %w(blue spots) })
=> {:class=>["red", "stripes", "blue", "spots"]}
```

Perfect.

Telling everyone to use arrays is probably a bit heavy-handed so instead [there's a new `Traits::HTMLAttributes::Attributes` class](https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/256/files#diff-0568c723f10a24d0a3cd8d98a2dfa59ed9c5174103ad96dfcc342abba1b4b38eR8) that has the responsibility of doing the conversion to arrays, the merging, and the conversion back to strings. We need to convert back to strings on the way out to get Rails 6.0.3 to pass, Rails 6.1.0+ will happily deal with values in either format.

### DIFF
--- a/lib/govuk_design_system_formbuilder/traits/html_attributes.rb
+++ b/lib/govuk_design_system_formbuilder/traits/html_attributes.rb
@@ -1,45 +1,50 @@
 module GOVUKDesignSystemFormBuilder
   module Traits
     module HTMLAttributes
-      class Parser
-        # target any element where the content is a string that represents a
-        # list of things separated by a space, e.g.,
-        #
-        # <span class="red spots">xyz</span>
-        #
-        # could be respresented as
-        #
-        # = span(class: %w(red spots)) { "xyz" }
-        TARGETS = [%i(class), %i(aria describedby)].freeze
-
-        def initialize(attributes)
-          @attributes = parse(attributes)
+      # Attributes eases working with default and custom attributes by
+      # * deeply merging them so both the default (required) attributes are
+      #   present
+      # * joins the arrays into strings to maintain Rails 6.0.3 compatibility
+      class Attributes
+        def initialize(defaults, custom)
+          @merged = defaults.deeper_merge(deep_split_values(custom))
         end
 
         def to_h
-          @attributes
+          deep_join_values(@merged)
         end
 
       private
 
-        def parse(attributes)
-          attributes.tap do |a|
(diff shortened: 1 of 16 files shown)


lib/govuk_design_system_formbuilder/traits/html_attributes.rb
```diff
@@ -1,8 +1,45 @@
...
+        end
+
+      private
+
+        def parse(attributes)
+          attributes.tap do |a|
+            a[:class]              = a[:class].split              if has_class_string?(a)
+            a[:aria][:describedby] = a[:aria][:describedby].split if has_aria_describedby_string?(attributes)
+          end
+        end
```
peteryates, line 30: The idea here is to use `TARGETS` to do do this dynamically.

peteryates:
Seeing as this approach totally fails on Rails 6.0.3 let's park it for the time being. I'll make an update to the guide instructing people to supply arrays instead of strings instead.

lib/govuk_design_system_formbuilder/traits/html_attributes.rb
```diff
@@ -1,8 +1,45 @@
 module GOVUKDesignSystemFormBuilder
   module Traits
     module HTMLAttributes
+      class Parser
+        # target any element where the content is a string that represents a
+        # list of things separated by a space, e.g.,
+        #
+        # <span class="red spots">xyz</span>
+        #
+        # could be respresented as
```
cpjmcquillan: *represented

peteryates: this comment's no longer in, instead the class just works through all the attributes.

cpjmcquillan (approved):

PR #261, opened 2021-04-10
Make date field value setting more lenient

### LINKED ISSUE #259: Error when attempting to render `govuk_date_field` with `nil` Date object values
After upgrading to the latest version: `2.4.0`, when attempting to render a date field using `govuk_date_field` with `nil` Date object segment values the following error is thrown: 
```
IndexError:
  #   offset 3 too large for struct(size:3)
```
Whilst running this outside tests the form renders correctly as we have empty string values specified for the [segments](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/elements/date.rb#L75) thus the fallback to `attribute.try(:[], MULTIPARAMETER_KEY[segment])` is never attempted after `attribute.try(segment)`.
```
> attribute
> #<struct  day="", month="", year="">
```
In the specs however since the values are `nil`, `attribute.try(segment)` seems to fallback to `attribute.try(:[], MULTIPARAMETER_KEY[segment])`, throwing the index error

```
> attribute
> #<struct  day=nil, month=nil, year=nil>
```

peteryates:
A recent change caused test failures in some projects because 'odd' values were used to set dates. These included a Struct with day, month and year accessors that allows nil values, as found in the Apply For Postgraduate Teacher Training project.

Now we're calling the method directly if the attribute responds to the segment; as this is now _within_ the if statement rather than a chain of ors, if the value is nil, nil will be returned.

There's an additional failure state to capture and inform when objectsthat are un-datelike are provided.

Fixes #259

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/date.rb
+++ b/lib/govuk_design_system_formbuilder/elements/date.rb
@@ -77,7 +77,11 @@ module GOVUKDesignSystemFormBuilder
         if attribute.respond_to?(segment)
           attribute.send(segment)
         elsif attribute.respond_to?(:fetch)
-          attribute.fetch(MULTIPARAMETER_KEY[segment])
+          attribute.fetch(MULTIPARAMETER_KEY[segment]) do
+            Rails.logger.warn("No key '#{segment}' found in MULTIPARAMETER_KEY hash. Expected to find #{MULTIPARAMETER_KEY.values}")
+
+            nil
+          end
         else
           fail(ArgumentError, "invalid Date-like object: must be a Date, Time, DateTime or Hash in MULTIPARAMETER_KEY format")
         end
--- a/lib/govuk_design_system_formbuilder/elements/date.rb
+++ b/lib/govuk_design_system_formbuilder/elements/date.rb
@@ -72,7 +72,15 @@ module GOVUKDesignSystemFormBuilder
       def value(segment)
         attribute = @builder.object.try(@attribute_name)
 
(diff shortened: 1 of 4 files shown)


peteryates:
☝🏽 All of Apply's tests pass now, except the `wkhtmltopdf` which always fails on my machine 😓

cpjmcquillan (approved):
LGTM

lib/govuk_design_system_formbuilder/elements/date.rb
```diff
@@ -72,7 +72,15 @@ def date_part(segment, width:, link_errors: false)
       def value(segment)
         attribute = @builder.object.try(@attribute_name)
 
-        attribute.try(segment) || attribute.try(:[], MULTIPARAMETER_KEY[segment])
+        return unless attribute
+
+        if attribute.respond_to?(segment)
+          attribute.send(segment)
+        elsif attribute.respond_to?(:fetch)
+          attribute.fetch(MULTIPARAMETER_KEY[segment])
```
cwrw, line 80: I'm wondering if using `fetch` is too strict, if someone defines a multi-parameter attribute that does not include the correct segment do we want to ignore it and return `nil` instead or throw a `key not found (KeyError)`?

peteryates: Personally I'd _want_ failure here, but I can see how everyone might not like that.

I amended the behaviour a bit so it'll fire a warning message if nothing's fetched but will return `nil`.

I'm going to add a subsequent PR that wraps `Rails.logger.warn` so they can be toggled on/off library-wide.

PR #263, opened 2021-04-12
Add two-thirds custom width option in text-field guide

david-mears-dfe:
I don't see a principled reason to skip this custom width in the guide. I am able to add the relevant width class (.govuk-!-width-two-thirds) successfully when I use the width attribute.


peteryates:
Definitely no principled reason, just me forgetting to add it. Thanks!

Issue #266, opened 2021-04-16
Date Handling Errors


eloisetait3:
We're seeing the error `invalid Date-like object: must be a Date, Time, DateTime or Hash in MULTIPARAMETER_KEY format` 
when inputting invalid dates to our form - both in browser and in tests. 

To recreate this - have a 3 field date input and insert incorrect values for example Day: 1 - Month: 1 - Year: 20201

This worked previously in 2.4.0, is this expected in 2.5.0, or are we expected to change our error handling?

peteryates:
Yeah - I believe this is due to you using the [GovUkDateFields gem](https://github.com/ministryofjustice/gov_uk_date_fields), which [looks like it's using this FormDate class](https://github.com/ministryofjustice/gov_uk_date_fields/blob/master/lib/gov_uk_date_fields/form_date.rb) to represent dates. It's not responding to `#day`, `#month` or `#year` (instead `#dd`, `#mm`, `#yyyy`) so it's not treated like a Ruby `Date` object.

The form builder sticks very closely to Rails' defaults, the `1i`/`2i`/`3i` style params work with the inbuilt multi-parameter attributes to build a `Date`.

The options to resolve this are:

* add a config option to the form builder that _also_ checks the `attribute` for other accessors ([here](https://github.com/DFE-Digital/govuk_design_system_formbuilder/blob/master/lib/govuk_design_system_formbuilder/elements/date.rb#L77)), so it might be something like `config.custom_date_accessors = %i(dd mm yyyy)`
* patch `govuk_uk_date_fields` and add some `alias_method` to point `#day` at `#dd` etc
* rework your date fields so they're more in-line with Rails - totally appreciate this might be a massive task so I'm more than happy to help with either of the two above if poss 😅

peteryates:
I had a chat with @zheileman and he was surprised people were still using the GovUkDateFields library, especially in conjunction with the form builder. We are of the opinion that moving to the form builder's (i.e. Rails') approach for handling dates and date params will be easiest in the long run.

As I said above - if it's a big task I'm willing to add an extra clause to the `if` statement linked above that'll allow you to continue using the latest version of the gem while you update your validations; I'll await a response before taking the next step.

eloisetait3:
Hi Peter - strange as we definitely don't use the GovUkDateFields library - it's definitely not in our yarn packs or gemfile. The issues must be arriving elsewhere?

peteryates:
[...]

PR #271, opened 2021-04-24
Allow logging instead of failure on invalid dates

peteryates:
Add option to allow date values to log when invalid instead of failing.

Failure is a sensible choice for most projects but is likely to become a reason why those in the process of reworking their codebase can't stay up to date with the gem.

The new option, enable_log_on_invalid_date, is disabled by default so the behaviour of the library won't change uninvited.


cpjmcquillan (approved):

peteryates:
The issue that this PR was created to resolve (#266) has been sorted. Closing this for the time being, but if other people encounter problems we can easily reinstate it.

PR #282, opened 2021-05-27
Prevent wrapping of empty block content

jsugarman:
#### What
Prevent wrapping of empty block content

#### Why
The existing govuk_radio_button and govuk_check_box
wrap blocks in a div even when there is not content.

This causes a padding issue in use cases where the content
is conditionally rendered.

Example of use case
```ruby
- @a_role_collection.each_with_index do |role, idx|
          = f.govuk_check_box(:roles,
            role,
            label: { text: role.humanize },
            link_errors: idx.zero?) do

            - if some_conditional?
              = f.govuk_collection_radio_buttons :some_radio_options,
                [['Yes','true'],['No','false']],
                :last,
                :first,
                inline: true,
                legend: { text: 'my legend', size: 's' }
```

### DIFF
--- a/lib/govuk_design_system_formbuilder/traits/conditional.rb
+++ b/lib/govuk_design_system_formbuilder/traits/conditional.rb
@@ -8,8 +8,11 @@ module GOVUKDesignSystemFormBuilder
       end
 
       def wrap_conditional(block)
+        block_result = block.call
+        return if block_result.blank?
+
         tag.div(class: conditional_classes, id: conditional_id) do
-          capture { block.call }
+          capture { block_result }
         end
       end
     end
--- a/spec/govuk_design_system_formbuilder/builder/check_boxes/check_box_spec.rb
+++ b/spec/govuk_design_system_formbuilder/builder/check_boxes/check_box_spec.rb
@@ -179,6 +179,14 @@ describe GOVUKDesignSystemFormBuilder::FormBuilder do
             }
           )
         end
+
+        context 'with no content' do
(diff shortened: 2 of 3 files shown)


jsugarman:
Hi there @peteryates, 

Getting this error in the CI. Is there some setup I have missed?
```
Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag
```

lib/govuk_design_system_formbuilder/traits/conditional.rb
```diff
@@ -8,8 +8,11 @@ def conditional_id
       end
 
       def wrap_conditional(block)
+        block_result = block.call
+        return if block_result.nil? || block_result.eql?('')
```
cpjmcquillan, line 12: Minor one but how about...
```suggestion
        return if block_result.blank?
```
or...
```suggestion
        return unless block_result.present?
```

jsugarman: Thanks @cpjmcquillan. Does the form builder aim to be framework (rails/activesupport) agnostic? I had not used `blank?` or `present?` to avoid those dependencies but would use `blank?` otherwise.

jsugarman: I see it is already using `blank?` in other place actually so will tweak as above.

jsugarman:
fixed by https://github.com/DFE-Digital/govuk_design_system_formbuilder/pull/283

PR #283, opened 2021-05-28
Don't render any conditional content when block is empty

peteryates:
When either no block or an empty block (i.e., one that fails Rails' `#presence` check) is provided in the block passed to a radio button or check box no conditional content shouldn't be rendered and the data-attribute that's used to toggle it should be omitted.

This tweak moves the logic to earlier in the object's initialisation so that both the `@conditional_content` and `@conditional_id` will not be set when the block is blank.

Closes #281, #282

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/radios/fieldset_radio_button.rb
+++ b/lib/govuk_design_system_formbuilder/elements/radios/fieldset_radio_button.rb
@@ -19,8 +19,8 @@ module GOVUKDesignSystemFormBuilder
           @link_errors     = has_errors? && link_errors
           @html_attributes = kwargs
 
-          if block_given?
-            @conditional_content = wrap_conditional(block)
+          if (conditional_block_content = block_given? && block.call.presence)
+            @conditional_content = wrap_conditional(conditional_block_content)
             @conditional_id      = conditional_id
           end
         end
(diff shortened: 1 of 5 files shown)


spec/govuk_design_system_formbuilder/builder/check_boxes/check_box_spec.rb
```diff
@@ -181,6 +181,20 @@
         end
       end
 
+      context 'when an empty block is given' do
+        subject { builder.send(*args) { "" } }
```
jsugarman, line 185: 👍

lib/govuk_design_system_formbuilder/elements/radios/fieldset_radio_button.rb
```diff
@@ -19,8 +19,8 @@ def initialize(builder, object_name, attribute_name, value, label:, hint:, link_
           @link_errors     = has_errors? && link_errors
           @html_attributes = kwargs
 
-          if block_given?
-            @conditional_content = wrap_conditional(block)
+          if block_given? && (conditional_block_content = block.call.presence)
```
jsugarman, line 22: Is there a reason for the different conditional statement here and at line 24: of the `fieldset_check_box.rb`

```
if block_given? && (conditional_block_content = block.call.presence)
```

and 

```
if (conditional_block_content = block_given? && block.call.presence)
```

I also wonder if it could it be refactored to:

```
if (conditional_block_content = block&.call&.presence)
```

but have not tried

peteryates: 🤦🏽 thanks, fixed

jsugarman (approved):
👍  Many thanks

jrmhaig:
Thanks, this looks great. I'll close #281

cpjmcquillan (approved):

PR #284, opened 2021-05-29
Remove duplication between fieldset checkboxes and radios

peteryates:
These two classes act in a very-similar manner, much of the functionality is duplicate. As all fieldset items are conditional, we can safely remove the `Traits::Conditional` and just treat everything as a `FieldsetItem`.

This is the same approach taken by the text fields and `Traits::Input`.


peteryates:
Thanks for suggesting this @jrmhaig, @jsugarman 👍🏽

PR #300, opened 2021-06-30
Fix multiple captures

jsugarman:


jsugarman:
sorry, this was not meant to be opened yet

PR #301, opened 2021-07-01
Fix multiple captures and their impact on nested field error links

jsugarman:
### What
Fix multiple captures and their impact on nested field error links
 - prevent the overhead of render calls being made multiple times
 - fix the side-effect of incrementing nested field-error ids, thereby breaking links between the error summary and the fields

### Why
We have a fairly complex nested field structure for a few of our forms
and error linking using this formbuilder was not working, but works against 
this branch.

On investigation I found that partials for the nested fields were being Rendered multiple times.
This did not actually render the fields multiple times, which I still do not understand
why not, but did increment the id of the nested fields.

#### Cause
- **1. Fieldset class plus super class capture the called block**
Digging into the code I found that the `RadioButtonsFieldset` and `CheckBoxesFieldset` 
inherit from the `Base` class which captures the `block.call` themselves AND via calls to
`super()`.

  Because `super(arg1,arg2,etc)` is called, the block is passed implicitly
whether you supply it as an arg to super, `&block`, or not. This
results in the block being called and captured twice `capture { block.call }`,
once for the fieldset and once by superclass `Base`.

  In the case of a block being passed that renders a partial it
means that rails renders the partial twice which has the
side effect of incrementing the ids of fields.

  _fixed by removing the fieldset level capture and relying on inheritance_

- **2. fieldset_item block capture**
Ironically, out of a change we asked for, and you kindly supplied, there is an additional 
`capture { block.call }` in the `FieldsetItem` trait. which we are also dependent on. 

  _fixed by passing the captured content directly_

### Example

A simplified version with only one fieldset might look like this 

```
# my_nested_fields_partial
    = f.fields_for :details, @details do |ff|
        = ff.govuk_text_field :name
```

```
# my_form_partial
= f.govuk_radio_buttons_fieldset :provider, do
    if some_condition
      = render partial: 'my_nested_fields_partial' 
```

**The resulting html**
`my_nested_fields_partial` is called twice, but only displays content once??, but does increment field-error ids so:
 id of the name field is `provider-details-attributes-1-name-field-error` while the summary error generated for it correctly links to `provider-details-attributes-0-name-field-error`
 
### Real example
We have [fieldsets inside fieldsets plus conditional blocks](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/app/views/shared/providers/_form.html.haml) which causes exponential  id incrementation issues, so that our first nested field ends up with an id of 15 instead of 0, the second with an id of 31.

Screenshot 2021-07-01 at 08 41 25

#### Additional
Possibly problematic block captures exist in other classes that inherit from the `Base` class and could result in similar issues being experienced. `lib/govuk_design_system_formbuilder/elements/submit.rb` in particular. I have not changed this in this PR however as I have not experienced problems with this myself.

### DIFF
--- a/lib/govuk_design_system_formbuilder/containers/check_boxes_fieldset.rb
+++ b/lib/govuk_design_system_formbuilder/containers/check_boxes_fieldset.rb
@@ -5,6 +5,8 @@ module GOVUKDesignSystemFormBuilder
       include Traits::Hint
 
       def initialize(builder, object_name, attribute_name, hint:, legend:, caption:, small:, classes:, form_group:, multiple:, &block)
+        fail LocalJumpError, 'no block given' unless block_given?
+
         super(builder, object_name, attribute_name, &block)
 
         @legend        = legend
--- a/lib/govuk_design_system_formbuilder/containers/check_boxes_fieldset.rb
+++ b/lib/govuk_design_system_formbuilder/containers/check_boxes_fieldset.rb
@@ -14,7 +14,6 @@ module GOVUKDesignSystemFormBuilder
         @classes       = classes
         @form_group    = form_group
         @multiple      = multiple
-        @block_content = capture { block.call }
       end
 
       def html
--- a/spec/support/shared/shared_block_examples.rb
+++ b/spec/support/shared/shared_block_examples.rb
@@ -39,3 +39,29 @@ shared_examples 'a field that accepts arbitrary blocks of HTML' do
     end
   end
 end
+
+shared_examples 'a fieldset that expects arbitrary blocks of HTML' do
+  context 'when no block is supplied' do
+    subject { builder.send(*args) }
+
+    specify { expect { subject }.to raise_error(LocalJumpError, /no block given/) }
+  end
+
+  context 'when block is supplied' do
+    subject { builder.send(*args, &example_block) }
+
+    let(:example_block) { proc { '<b>some content</b>' } }
+
+    specify { expect { subject }.not_to raise_error }
+    specify { expect { |b| builder.send(*args, &b) }.to yield_control.once }
+
+    context 'with render' do
+      let(:example_block) { proc { helper.render(html: '<b>rendered</b>') } }
+
+      before { allow(helper).to receive(:render) }
+      before { subject }
+
+      specify { expect(helper).to have_received(:render).once }
+    end
+  end
+end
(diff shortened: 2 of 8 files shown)


jsugarman:
Hi @peteryates, hopefully this PR makes sense. happy to discuss here or via slack, as usual

peteryates:
Thanks @jsugarman! Yeah, this looks good to me and your solution is definitely easier to follow. Will take a closer look later and get this merged and released.

spec/support/shared/shared_block_examples.rb
```diff
@@ -39,3 +39,31 @@
...
+
+    context 'with render' do
+      before { allow(helper).to receive(:render) }
+
+      let(:example_block) { proc { helper.render(html: '<b>rendered</b>') } }
+
+      specify do
+        subject
+        expect(helper).to have_received(:render).once
+      end
```
peteryates, line 66: I think reordering the setup makes this a tiny bit easier to follow

```suggestion
      let(:example_block) { proc { helper.render(html: '<b>rendered</b>') } }
      
      before { allow(helper).to receive(:render) }
      before { subject }

      specify { expect(helper).to have_received(:render).once }
```

lib/govuk_design_system_formbuilder/containers/check_boxes_fieldset.rb
```diff
@@ -5,6 +5,8 @@ class CheckBoxesFieldset < Base
       include Traits::Hint
 
       def initialize(builder, object_name, attribute_name, hint:, legend:, caption:, small:, classes:, form_group:, multiple:, &block)
+        raise LocalJumpError, 'no block given' unless block_given?
```
peteryates, line 8: Immediately erroring here is a huge improvement. I tend to use `fail` instead of `raise` where there's no intention of rescuing.

spec/support/shared/shared_block_examples.rb
jsugarman: Do you prefer to keep the befores separate?

or 
```
before do
  allow(helper).to receive(:render)
  subject
end

jsugarman: have applied as separate befores for now as above

peteryates (approved):
Definite improvement 👍🏽

jsugarman:
Great, many thanks @peteryates . I assume you have to merge, especially as the test coverage check always fails when the PR is from a fork.

PR #310, opened 2021-08-24
Enable error presenter injection

jsugarman:
#### What
Enable summary error customisation

#### Why
Our use case currently requires us to present summary errors for multiple nested fields sets with wording to indicate the nested field set's _index and error_. This is an accessibility issue to some degree too.

**Example use case**
Users are able to add one or many nested objects, suppliers in this example, on a single form. The field level error is expected to display the "short" error message for the attribute while the summary message must display the error for the attribute as well as a plain english index of the nested field object. 

Screenshot 2021-08-24 at 16 42 23

_We also have nested fields within nested fields in the application which behave similarly._ 

#### How
We have a large set of validations that work in conjunction with a presenter (Decorator pattern) and i18n translations files to achieve this customised behaviour. The [presenter and dependencies can be seen here, for reference](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/tree/master/app/presenters/error_message).

This PR aims to enable us to inject this presenter to achieve the same behaviour using the govuk-formbuilder. This works by simply relying on duck-typing of an injected presenter and its `summary_error_for` method to retrieve the custom "ordinalized" message

### DIFF
--- a/guide/content/introduction/error-handling.slim
+++ b/guide/content/introduction/error-handling.slim
@@ -44,4 +44,27 @@ p.govuk-body
       summary link should take the user to the first field. As the form builder
       doesn't know the order in which fields will be rendered, it must be specified.
 
+== render('/partials/example-fig.*',
+  caption: "Custom summary error presenter injection",
+  code: form_with_presenter_injection,
+  sample_data: departments_data_raw,
+  show_errors: :presenters,
+  hide_data: true,
+  hide_html_output: true) do
+
+  p.govuk-body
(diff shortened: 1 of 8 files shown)


jsugarman:
our related app PR for reference https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4164

peteryates:
This all looks great.

My one concern would be that the functionality isn't really _discoverable_. Would it make sense to create a default presenter class within the library that implements something along the lines of:

```
class GOVUKDesignSystemFormBuilder::ErrorSummaryPresenter
  # some useful guidance on how to write an error summary
  def message_for(attribute)
     fail(NoMessageForMethodImplemented, "you need to override :message_for with your custom error format")
  end
end
```

Then you can inherit and override it with your custom presenter?

jsugarman:
closing in favour of https://github.com/DFE-Digital/govuk-formbuilder/pull/312

PR #311, opened 2021-08-28
Add a default error summary presenter

peteryates:
The intention of this presenter is that it provides the current basic behaviour but also provides a convenient way for people with specific requirements to inject them.

It currently requires that it's initialized with an error_messages hash but we may yet tweak that so an instance can be passed in and the error messages set on it.

Refs #310

### DIFF
--- a/lib/govuk_design_system_formbuilder/presenters/error_summary.rb
+++ b/lib/govuk_design_system_formbuilder/presenters/error_summary.rb
@@ -1,0 +1,27 @@
+module Presenters
+  # This is the default presenter for {GOVUKDesignSystemFormBuilder::Elements::ErrorSummary} and is
+  # intended to be easily replaceable should you have specific requirements that aren't met here.
+  #
+  # The basic behaviour is to always show the first error message. In Rails, error message order is
+  # determined by the order in which the validations run, but if you need to do any other transformation
+  # or concatenation, this is the place to do it.
+  class ErrorSummaryPresenter
+    # @param [Hash] error_messages the error message hash in a format that matches Rails'
+    #   `object.errors.messages`, so the format should be:
+    #
+    # @example Input format:
+    #   ErrorSummaryPresenter.new({ attribute_one: ["first error", "second error"], attribute_two: ["third error"] })
+    def initialize(error_messages)
+      @error_messages = error_messages
+    end
+
+    # Converts +@error_messages+ into an array of argument arrays that will be
+    # passed into {GOVUKDesignSystemFormBuilder::Elements::ErrorSummary#list_item}.
+    #
+    # @example Output format given the input above:
+    #   [[:attribute_one, "first error"], [:attribute_two, "third error"]]
+    def formatted_error_messages
+      @error_messages.map { |attribute, messages| [attribute, messages.first] }
+    end
+  end
+end
--- a/lib/govuk_design_system_formbuilder/presenters/error_summary.rb
+++ b/lib/govuk_design_system_formbuilder/presenters/error_summary.rb
@@ -18,6 +18,8 @@ module Presenters
     # Converts +@error_messages+ into an array of argument arrays that will be
     # passed into {GOVUKDesignSystemFormBuilder::Elements::ErrorSummary#list_item}.
     #
+    # @return [Array<Array(Symbol, String)>] array of attribute and message arrays
+    #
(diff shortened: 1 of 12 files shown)


peteryates:
@jsugarman - having had a bit of a think about it, I _think_ this approach will give you what you need. If we always use a presenter it may as well handle the collection of error messages rather than one-at-a-time.

If you need more info, we could always optionally pass the object to the initializer too, something like:

```ruby
def presenter
  @presenter.new(error_messages, @builder.object)
end
```

Alternatively, we could pass an instance of the presenter to `#govuk_error_summary`. Happy with any of the approaches.

jsugarman:
Many thanks @peteryates. I just saw this when I pushed up my new branch directly against this repo. 

I will take a better look when I have time but my initial thoughts is that I don't think this approach will work for our custom error presenter without a major, and rather painful, rewrite on our end. We instantiate our presenter with the model object and a translation file (for lookups) and then have methods to retrieve the summary and field level errors. We pass custom attributes/keys which are then converted to ordinalize nested instances (e.g. attribute : defendant_2_first_name, message `blank` would become "Enter a first name for the second defendant"). 

I had added [some thoughts to this commit](https://github.com/DFE-Digital/govuk-formbuilder/commit/dfc62d09bae61b349123e53771f98bfb7d7546fe) while working.

However I am not even sure the approach I have taken in my PR would work for us going forward and applying this functionality to other models because I can see that we generate separate line item summary errors for the same attribute with multiple errors. 

I will take a better look at your PR when I can but do let me know any thoughts, problems, improvements on mine when you can.

peteryates:
@jsugarman - thanks for the feedback, I kind of had this in mind - I've updated it so it'll accept any object in addition to a presenter class.

I believe this will work with your PR now with minimal changes, I think you just need to add a `#formatted_error_messages` to your presenter class and have it loop through the object's error attributes yourself. I'll take a closer look when I get chance.

jsugarman:
I just  pulled your branch and had a quick play tested against an implementation of the `formatted_error_messages` in our presenter and it works 🎉 .

I will have a further look but it I like it simplicity

lib/govuk_design_system_formbuilder/presenters/error_summary.rb
```diff
@@ -0,0 +1,27 @@
...
+    #
+    # @example Input format:
+    #   ErrorSummaryPresenter.new({ attribute_one: ["first error", "second error"], attribute_two: ["third error"] })
+    def initialize(error_messages)
+      @error_messages = error_messages
+    end
+
+    # Converts +@error_messages+ into an array of argument arrays that will be
+    # passed into {GOVUKDesignSystemFormBuilder::Elements::ErrorSummary#list_item}.
+    #
```
jsugarman, line 20: Could add a technical definition of the return value. I notice you have this style in some technical documentation which i rather like. I am, however, not sure how an you would write the definition of an Array of Array elements that have a symbol followed by a string but here is my stab at it
```
# @return [Array<Array<Symbol, String>>] array of attribute and message arrays
```

jsugarman (approved):
🥇 Have this working as I'd hope for against our [app branch](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/pull/4164). 

Couple of minor documentation comments plus do you think this is worth its own entry in the guide to provide an interactive example. You could [cherry pick and amend this commit](https://github.com/DFE-Digital/govuk-formbuilder/pull/312/commits/0ffcc0bb34e896910f4acb56db2117221e7e08c9), or am happy to add myself.

lib/govuk_design_system_formbuilder/presenters/error_summary.rb
peteryates: I always have to have a play around with this in YARD to get it working, will take a closer look

lib/govuk_design_system_formbuilder/presenters/error_summary.rb
```diff
@@ -18,6 +18,8 @@ def initialize(error_messages)
     # Converts +@error_messages+ into an array of argument arrays that will be
     # passed into {GOVUKDesignSystemFormBuilder::Elements::ErrorSummary#list_item}.
     #
+    # @return [Array<Array(Symbol, String)>] array of attribute and message arrays
```
jsugarman, line 21: 👀  👍

PR #313, opened 2021-08-31
Improve the setup of error presenters

peteryates:
Now we're checking that both the custom class and instance implements `#formatted_error_messages` while simplifying the object setup.

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/error_summary.rb
+++ b/lib/govuk_design_system_formbuilder/elements/error_summary.rb
@@ -11,7 +11,7 @@ module GOVUKDesignSystemFormBuilder
         @link_base_errors_to = link_base_errors_to
         @html_attributes     = kwargs
         @order               = order
-        @presenter           = setup_presenter(presenter)
+        @presenter           = presenter
       end
 
       def html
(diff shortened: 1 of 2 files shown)


spec/govuk_design_system_formbuilder/builder/error_summary_spec.rb
```diff
@@ -472,6 +472,15 @@ def formatted_error_messages
...
+          let(:non_presenter) { OpenStruct }
+          subject { builder.send(*args, presenter: non_presenter) }
+
+          specify "fails with an appropriate error message" do
+            expect { subject }.to raise_error(ArgumentError, "error summary presenter doesn't implement #formatted_error_messages")
+          end
+        end
       end
 
       context "as an instance" do
```
jsugarman, line 486: nit - not related so ignore as you like but - rubocop-rspec perfers `with/without/when`
```suggestion
      context "with an instance" do
```

peteryates: yeah - i disagree with Rubocop in this instance (and several others!) 😄

lib/govuk_design_system_formbuilder/elements/error_summary.rb
```diff
@@ -34,15 +34,15 @@ def summary
...
         end
       end
 
       # If the provided @presenter is a class, instantiate it with the sorted
       # error_messages from our object. Otherwise (if it's any other object),
       # treat it like a presenter
-      def setup_presenter(presenter)
-        (presenter.is_a?(Class) ? presenter.new(error_messages) : presenter).tap do |p|
+      def presenter
+        (@presenter.is_a?(Class) ? @presenter.new(error_messages) : @presenter).tap do |p|
```
jsugarman: 👍 🎖️

PR #320, opened 2021-10-04
Switch from list of unmergeable to mergeable attrs

### LINKED ISSUE #318: Setting a JSON object as an HTML attribute on form elements doesn't work
Creating a radio button using something like:

```
<% object = {"some_key"=>["Yes - a regular", "Yes - a reserve"], "another_key"=>["Yes - a regular", "Yes - a reserve"]} %>

<%= f.govuk_radio_button question_key, val, label: { text: val },
            "data-controller": "stimulus-controller",
            "data-action": "click->stimulus-controller#method",
            "data-info": object.to_json
        %>
```

Means that here https://github.com/DFE-Digital/govuk-formbuilder/blob/master/lib/govuk_design_system_formbuilder/traits/html_attributes.rb#L27 Form Builder treats the json value as a string to split and duplicate words are removed. This means that instead of correctly outputting the following HTML:

```
data-info="{&quot;some_key&quot;:[&quot;Yes - a regular&quot;,&quot;Yes - a reserve&quot;],&quot;another_key&quot;:[&quot;Yes - a regular&quot;,&quot;Yes - a reserve&quot;]}
```

You get

```
data-info="{&quot;some_key&quot;:[&quot;Yes - a regular&quot;,&quot;Yes reserve&quot;],&quot;another_key&quot;:[&quot;Yes reserve&quot;]}
```

(Note. "Yes - a reserve" has been turned into "Yes reserve").

You can fix this by modifying `Attributes#deep_split_values` to something like:

```
when String
                            begin
                              JSON.parse(value)
                              value
                            rescue JSON::ParserError => e
                              split_mergeable(key, value, parent)
                            end
```

But that seems likely to break other use cases? Is there another supported way to do this currently or would a change to enable it be viable?

peteryates:
Previously we held a list of HTML attributes that should never be merged together by this library. It's clear that the opposite approach makes more sense, there's only a small(ish) number of HTML attributes that we should try - those that allow a space-separated list of items.

The common examples are class, aria-describedby and aria-labelledby, but there are [quite a few other aria ones](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#state_prop_def).

Fixes #318


baarkerlounger:
Can confirm this fixes my issue.

peteryates:
Glad to hear it. Also, huge thanks for your PR @baarkerlounger, much appreciated. Will get this released today.

PR #325, opened 2021-11-11
Allow URL overrides for error_summary entries

elliotcm:
This allows us to write deep links to other pages via the presenter.

Let me know if you'd like me to remove the tools-version file or do it differently, bundler wouldn't run without it set.


peteryates:
Thanks @elliotcm. I like the idea but rather than a hash I think it would be cleaner if `url` was a separate (optional) third argument. I had a play around with the idea in #326 - think this could work for you?

elliotcm:
Closed in favour of https://github.com/DFE-Digital/govuk-formbuilder/pull/326

PR #329, opened 2021-11-26
Support the maxlength attribute when building the Date element

willfish:
### What

This PR adds the ability for users of the govuk_date_field and Date element to
be able to specify a maxlength attribute on each of the inputs. It defaults to
4 chars for the year, 2 chars for the month and 2 chars for the day.

- [x] Updates the form interface method :govuk_date_field to take the param
       maxlength_enabled (defaulting to false) and pass it to the Elements::Date constructor
- [x] Updates the Date element to add the maxlength value to the day, month and year inputs if enabled
- [x] Adds test coverage for each of these cases
- [x] Adds documentation and examples

### Why

We want to be able to control the number of characters the user is able
to input for the day, month and year inputs without hitting our controller
endpoint and having validations kick in.

### AC

- [x] Does not change existing behaviour

### DIFF
--- a/guide/content/form-elements/date-field.slim
+++ b/guide/content/form-elements/date-field.slim
@@ -63,4 +63,15 @@ section
         should be taken into account when displaying the data or using it for
         analysis.
 
+  == render('/partials/example-fig.*',
+    caption: 'Specifying client side length restrictions on date inputs with maxlength',
+    code: maxlength_enabled_field,
+    hide_html_output: true) do
+
+    p.govuk-body
+      | When desiring client side restrictions on the number of characters each
+        input requires, you can use the <code>maxlength_enabled</code> parameter.
+        This will default the day and month inputs to 2 characters and the year
+        input to 4 characters.
+
 == render('/partials/related-info.*', links: date_info)
--- a/guide/content/form-elements/date-field.slim
+++ b/guide/content/form-elements/date-field.slim
@@ -69,7 +69,7 @@ section
     hide_html_output: true) do
 
     p.govuk-body
-      | When desiring client side restrictions on the number of characters each
+      | To enable client side restrictions on the number of characters each
         input requires, you can use the <code>maxlength_enabled</code> parameter.
         This will default the day and month inputs to 2 characters and the year
         input to 4 characters.
--- a/lib/govuk_design_system_formbuilder/elements/date.rb
+++ b/lib/govuk_design_system_formbuilder/elements/date.rb
@@ -56,33 +56,21 @@ module GOVUKDesignSystemFormBuilder
       def day
         return if omit_day?
 
-        if maxlength_enabled?
-          date_part(:day, width: 2, link_errors: true, maxlength: 2)
-        else
-          date_part(:day, width: 2, link_errors: true)
-        end
+        date_part(:day, width: 2, link_errors: true)
       end
 
       def month
-        if maxlength_enabled?
-          date_part(:month, width: 2, link_errors: omit_day?, maxlength: 2)
-        else
-          date_part(:month, width: 2, link_errors: omit_day?)
-        end
+        date_part(:month, width: 2, link_errors: omit_day?)
       end
 
       def year
-        if maxlength_enabled?
-          date_part(:year, width: 4, maxlength: 4)
-        else
-          date_part(:year, width: 4)
-        end
+        date_part(:year, width: 4)
       end
 
-      def date_part(segment, width:, link_errors: false, maxlength: nil)
+      def date_part(segment, width:, link_errors: false)
         tag.div(class: %(#{brand}-date-input__item)) do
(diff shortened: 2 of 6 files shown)


lib/govuk_design_system_formbuilder/elements/date.rb
```diff
@@ -105,7 +122,8 @@ def input(segment, link_errors, width, value)
           pattern: pattern(segment),
           inputmode: 'numeric',
           value: value,
-          autocomplete: date_of_birth_autocomplete_value(segment)
+          autocomplete: date_of_birth_autocomplete_value(segment),
+          maxlength: maxlength,
```
peteryates, line 126: I think the `if`/`else` statements could be avoided by adding a `MAX_LENGTHS` constant, along the lines of

```
MAX_LENGTHS = { day: 2, month: 2, year: 4 }.freeze
```

then here if it falls back to `nil` the attribute will automatically be omitted by Rails:

```suggestion
          maxlength: max_length(segment),
        end
        
        def max_length(segment)
          return unless maxlength_enabled?
          
          MAX_LENGTHS.fetch(segment)
        end
```

guide/content/form-elements/date-field.slim
```diff
@@ -63,4 +63,15 @@ section
...
+  == render('/partials/example-fig.*',
+    caption: 'Specifying client side length restrictions on date inputs with maxlength',
+    code: maxlength_enabled_field,
+    hide_html_output: true) do
+
+    p.govuk-body
+      | When desiring client side restrictions on the number of characters each
+        input requires, you can use the <code>maxlength_enabled</code> parameter.
+        This will default the day and month inputs to 2 characters and the year
+        input to 4 characters.
```
peteryates, line 75: "When desiring" doesn't sound quite right here - maybe "To enable client side restrictions"?

peteryates (commented):
Thank you @willfish - extra thanks for updating the guide too! This looks great, I left a couple of minor comments but overall it's a definite improvement.

lib/govuk_design_system_formbuilder/elements/date.rb
peteryates, line 126: As we're already passing in `width` here for setting the input width I think we can just use it directly. Doing something like this will allow you to avoid having the `if`/`else` statements in `#day`, `#month` and `#year`.

```suggestion
          maxlength: (width if maxlength_enabled?)
```

willfish:
No worries @peteryates. Appreciate the speedy review and feedback points. I'll get on these this morning and push some changes :).

lib/govuk_design_system_formbuilder/elements/date.rb
willfish: This is much cleaner :pray:

willfish:
Think that's much tidier now :). Let me know if you want any other changes @peteryates

peteryates (approved):
Looks great!

willfish:
The failure looks like a configuration error @peteryates:
```
 you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag
```

peteryates:
Ah yeah - it's because the PR is external it doesn't have access to the environment variable set in this repo. I'll grant you access so you can raise your next PR directly!

Thanks.

willfish:
@peteryates Any chance of cutting a release for this? I can't seem to find 2.8.0 on rubygems :). Also, I accepted the invite - thanks for sorting that.

peteryates:
Yeah - should go out today along with #330

willfish:
Thanks @peteryates.

edwardhorsford:
A word of caution that in many cases, maxlength on an inputs doesn't help, and you're pretty much always better off with server based validation.

The key issue is that browsers don't announce maxlength, and there's no indication to users that truncation has happened silently. Sighted users *might* notice, but a user of a screen reader wouldn't. Additionally, as date inputs are already very short, it's even more likely a sighted user wouldn't notice the truncation.

It's for this reason it's not a recommended attribute in the design system - there's no practical benefits, and several downsides / pitfalls. In the design system you could add it via adding attributes, but it's not specifically supported.

willfish:
@edwardhorsford I really appreciate your comment and will feed it back as part of the QA process that led to this change. I guess it's okay having this in the gem if users strictly want it. I'll be reviewing our use of maxlength globally as a result of this comment in an upcoming sprint

edwardhorsford:
@willfish no worries!

On the face of it it seems useful, but because of how browser handle it (or not, really), it's basically all downsides for users. 

For instance I've been tripped up several times by sites that use maxLength on their password creation forms to silently truncate input - so you type a long password and never realise that the one they've saved is *shorter* than what you typed. No validation to tell you you've chosen something too long. If the login form also truncates this may be ok - but if it doesn't, it means the password you know (and have in a password manager) doesn't actually match what was stored.

PR #334, opened 2021-12-18
Version 3.0.0

peteryates:
This release updates the library to work with the [GOV.UK Design System v4.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.0.0) and [Rails 7](https://rubyonrails.org/2021/12/15/Rails-7-fulfilling-a-vision).

It contains breaking changes and drops support for Rails 6.0.X and Ruby 2.6.0.X.


guide/content/stylesheets/components/_example.sass
```diff
@@ -32,5 +32,5 @@
 .final-decision-warning
   @extend %govuk-body-s
 
-  background: govuk-colour("pink")
-  padding: govuk-spacing(2)
\ No newline at end of file
+  background: govuk-tint(govuk-colour('pink'), 92%)
+  padding: govuk-spacing(2)
```
paulrobertlloyd, line 36: Ah yes. I was having trouble reloading my styles locally. Was thinking of having more obvious styling for these classes, using background colours and contrasting text. So this would have a pink background and white text.

peteryates: Ah I noticed the styles not refreshing too, will take a look now.

peteryates: I've given you access to this repo, please feel free to commit or raise PRs directly

peteryates: Tried white on the GOV.UK shade of pink but that didn't have necessary contrast either. Live reload should be working again now.

I'm going to make the Components version of the guide reflect all these changes and then  when the dust has settled look at standardising things (maybe the styles into a npm package) to try and reduce any duplication.

paulrobertlloyd:
Have updated the example styles to all use background colours (I wasn’t seeing the text decoration on Safari, even when I added prefixed properties, and I think backgrounds stand out better).

Have also updated the line length/code block widths as suggested, and tweaked some spacing and heading sizes.

Also reinstated copy fonts and images into the assets folder… not sure what issue you were seeing?

paulrobertlloyd:
Also updated the content on the supported versions page, using a table to show the support matrix; I think this works better, but feel to revert/tweak.

peteryates:
> Also reinstated copy fonts and images into the assets folder… not sure what issue you were seeing?

When it's `cp -r node_modules/govuk-frontend/govuk/assets/ content/assets/` I end up with the structure `content/assets/assetss`:

```
❯ ls --tree output
output
├── assets
│  ├── assets
│  │  ├── fonts
│  │  │  ├── bold-affa96571d-v2.woff
│  │  │  ├── bold-b542beb274-v2.woff2
```

paulrobertlloyd:
Hmmm, I end up with:

```
output
├── assets
├── fonts
│  ├── bold-affa96571d-v2.woff
│  ├── bold-b542beb274-v2.woff2
```

What system are you using? Is it Windows? I wonder if we need something a bit more robust, like an npm package that is system-independent?

peteryates:
Linux here, I was wondering if it might be platform specific. Will take a closer look tomorrow.

peteryates:
I believe I fixed the copying issue, tested it out (in a dry run, not with this actual project) on my old Macbook and it looks ok.

New graphic on the homepage is amazing! 💎

paulrobertlloyd:
> I believe I fixed the copying issue, tested it out (in a dry run, not with this actual project) on my old Macbook and it looks ok.

Yup, works for me! 🚀

PR #339, opened 2021-12-29
Update documentation design

paulrobertlloyd:
I saw https://github.com/DFE-Digital/govuk-components/pull/279 and got distracted…

* updates homepage content and illustration
* uses a design more aligned to GOV.UK design system (controversial, but plenty of other tools use the font and crown)
* refactors CSS into components
* uses simpler and more consistent Slim markup
* fixes manual line wrapping to meet 80 characters
* uses smart quotes
* combines rendered and HTML output into tab UI
* renames fields to use same names as used in the GOV.UK Design System docs
* copies `govuk-frontend` assets, styles and JavaScript (using `postinstall` in `package.json` – is there a way to do this with Nanoc?)

Sorry this is one big PR, but I got carried away 😊 

New homepage design


paulrobertlloyd:
A few thoughts having gone through this exercise…

* Can I just register my long held hatred of whitespace dependent syntax. HAML, Slim, Sass, CoffeeScript, Pug… absolutely awful and pointless exercises. Made updating the guide much harder as had to relearn/figure out how HTML and Ruby is represented in this alternate bizarro universe where using a curly brace or semicolon is somehow considered a sin 🤬 `</rant>`
* Is there a better option than Nanoc for the guide? It’d be much easier to write/update Markdown content files rather than write Slim. You’d ideally want a generator that allowed you to insert Ruby partials.
* Related, is there a generator that would allow you to use the GOV.UK components gem for many of these partials? Maybe this is possible with Nanoc?

peteryates:
Thank you so much for this ❤️ 

I'm totally open to improving the docs and making it easier for people to contribute. Even though I love _proper_ Sass syntax I understand it's marmiteesque, I'll switch that over to SCSS.

Originally, way back, I tried various static site generators and Nanoc (which is more of a build-your-own SSG kit really) was the only one that made it straightforward for me to integrate the library directly. I do hear you though, and will take a closer look at making it more straightforward. Markdown's already [a first class citizen in Nanoc](https://nanoc.app/doc/tutorial/#write-pages-in-markdown), but as you say, we just need a way of injecting partials.

Having had a quick look, the one downside to using a three-quarters column is that the code blocks often get chopped and need a bit of sideways scrolling. Perhaps they can keep their full width but the other elements are limited to 75% - something like this maybe?

Screenshot from 2021-12-29 18-44-53

Either way, huge thanks for the effort. Massive improvements (and no more shady symlinks 😂).

PR #349, opened 2022-02-10
Remove tabindex from Error Summary

frankieroberto:
The `tabindex` attribute on the Error Summary no longer needs to be set within the HTML, as this is set via javascript, as of govuk-frontend 4.0.1.

See [govuk-frontend 4.0.1 release notes](https://github.com/alphagov/govuk-frontend/releases/tag/v4.0.1).


peteryates (approved):
Ah good catch. Thanks @frankieroberto

PR #374, opened 2022-07-17
Add config to prevent automatic boldening of radio button labels in a collection

### LINKED ISSUE #373: Labels in radio button collections that include hint text are unexpectedly emboldened
The form builder currently has opinionated behaviour with regards to [radio button collections that include hint text](https://govuk-form-builder.netlify.app/form-elements/radios/#radios-collection-with-descriptions):

> When descriptions are enabled in a radio button collection the labels are made **bold** by default. This provides additional contrast and makes the labels stand out from the hint.

## Issues with this behaviour

* **It’s unexpected:** adding hint text shouldn’t change the appearance of a corresponding label (not least because the same behaviour isn’t exhibited in the Nunjucks components provided by GDS)
* **It’s inconsistent:** this behaviour is only exhibited when using `govuk_collection_radio_buttons`, not `govuk_radio_buttons_fieldset`. The equivalent `govuk_collection_check_boxes` helper also doesn’t embolden label options with hint text
* **It’s not a recommendation:** The GOV.UK Design System makes no such recommendation, and [its examples don’t show emboldened labels](https://design-system.service.gov.uk/components/radios/#radio-items-with-hints).
* **It’s not documented:** while the behaviour is noted, the documentation doesn’t mention how to disable this behaviour (by using `bold_labels: false`)
* **It requires additional code:** to get the expected behaviour requires adding `bold_labels: false` to every radio collection
* **It cannot be disabled globally:** each collection has to have this behaviour disabled individually, leading to potential inconsistencies across different forms

## Resolution

1. Document the `bold_labels` option on `govuk_collection_radio_buttons`
2. Add a global config option for this feature. By default this option should maintain the current behaviour. Setting this option to `false` would disable this behaviour across all radio collections (except those which have `bold_labels: true`, of course)
3. Consider including a breaking change in the 4.0.0 release. 2 possible options:
    1. Remove this behaviour entirely
    2. Keep the global option, but change its value so that this behaviour is disabled by default

peteryates:
The original behaviour of the form builder followed the examples in the Design System documentation at the time and emboldened any radio button labels when hints were present.

[These examples were dropped in 2020](https://github.com/alphagov/govuk-design-system/pull/1284) and the guidance changed. In the referenced PR I said I'd look into changing the behaviour in the form builder but never did.

Modifying the default would be a breaking change so we're just adding a new configuration option, `default_collection_radio_buttons_auto_bold_labels`, for the time being. The default will remain `true` for the time being but may change at the next major release.

To disable the automatic emboldening behaviour add this line to your configuration:

```ruby
GOVUKDesignSystemFormBuilder.configure do |c|
  c.default_collection_radio_buttons_auto_bold_labels = false
end
```

Fixes #373

### DIFF
--- a/guide/content/form-elements/radios.slim
+++ b/guide/content/form-elements/radios.slim
@@ -39,9 +39,11 @@ p.govuk-body
       null or empty no hint will be rendered.
 
   p.govuk-body
-    | When descriptions are enabled in a radio button collection the labels are
-      made <strong>bold</strong> by default. This provides additional contrast
-      and makes the labels stand out from the hint.
+    | When descriptions are enabled in a radio button collection the labels are made
+      <strong>bold</strong> by default. This makes the labels stand out from the hints.
+      The auto bolding behaviour can be disabled by changing the
+      <code>default_collection_radio_buttons_auto_bold_labels</code>
+      configuration flag to <code>false</code>.
 
 == render('/partials/example.*',
   caption: 'Small radios',
--- a/lib/govuk_design_system_formbuilder.rb
+++ b/lib/govuk_design_system_formbuilder.rb
@@ -43,6 +43,11 @@ module GOVUKDesignSystemFormBuilder
   # * +:default_collection_radio_buttons_include_hidden+ controls whether or not
   #   a hidden field is added when rendering a collection of radio buttons
   #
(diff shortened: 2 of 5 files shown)


tvararu (approved):
Nice work!

paulrobertlloyd:
Brilliant, thanks!

PR #376, opened 2022-07-20
Ensure date field without a day works correctly

thomasleese:
In https://github.com/DFE-Digital/govuk-formbuilder/pull/57 the date field was modified to support showing an approximate date (without a day field). However, it seems like it's not quite working as expected as Rails ignores the field as there's no HTML `input` element for the day part of the date.

This is demonstrated by https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/250/commits/03e6281e25081610aa70a9b22b541dac256b6c4a which required me to manually add hidden input fields for the two approximate date fields representing the day component, and setting the value to "1".

The built-in Rails equivalent date select helper also renders a hidden field for the day component, so it makes sense that we match this behaviour: https://github.com/rails/rails/blob/b5a758db1ba16233f53c76dea588c0e3853c7c42/actionview/lib/action_view/helpers/date_helper.rb#L814-L820

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/date.rb
+++ b/lib/govuk_design_system_formbuilder/elements/date.rb
@@ -53,7 +53,14 @@ module GOVUKDesignSystemFormBuilder
       end
 
       def day
-        return if omit_day?
+        if omit_day?
+          return tag.input(
+            id: id(:day, false),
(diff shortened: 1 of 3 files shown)


peteryates (approved):
Looks great!

PR #386, opened 2022-09-12
Improve the localisation of nested fields

peteryates:
Experimental branch with some improvements to nested localisation

Co-authored-by: zheileman <zheileman@users.noreply.github.com>


peteryates:
@zheileman - how's this?

```ruby
> "steps_case_codefendants_form[codefendants_attributes][0][foobar_attributes][1]".scan(/[[:alpha:]_]+/)

=> ["steps_case_codefendants_form", "codefendants_attributes", "foobar_attributes"]
```

zheileman:
@peteryates Sorry got distracted yesterday didn't look into this anymore. I think that's perfect. It might fail if there are numbers in any part of the `@object_name` other than the indices, I'm thinking for example a versioned form object might have the following namespace:

```ruby
module Steps
  module Case
    module V2
      class CodefendantsForm
      end
    end
  end
end
```

Producing the following:

```ruby
> "steps_case_v2_codefendants_form[codefendants_attributes][0]".scan(/[[:alpha:]_]+/)
=> ["steps_case_v", "_codefendants_form", "codefendants_attributes"]
```

I'm not sure how common this might be. I know I've used this in the past at least once.

I came up with this other alternative that I think will preserve numbers anywhere other than in the brackets (indices).
I tried to do this in one single regex but couldn't figure out, so I tried splitting into two (I'm sure there are better ways of doing this):

```ruby
> "steps_case_v2_codefendants_form[codefendants_attributes][0][foobar_attributes][1]".scan(/[\w]+/).grep(/[^[:digit:]]/)
=> ["steps_case_v2_codefendants_form", "codefendants_attributes", "foobar_attributes"]
```

Not sure if there are some other scenarios we might need to consider 🤔

peteryates:
cc @zheileman 

Some specs covering some path combinations with a regex: 6ca6657e

zheileman:
@peteryates I think that covers all possible naming and then some more 😄 🥳

lib/govuk_design_system_formbuilder/traits/localisation.rb
```diff
@@ -1,6 +1,10 @@
 module GOVUKDesignSystemFormBuilder
   module Traits
     module Localisation
+      # starts with an letter that is followed by other word characters or
+      # spaces, zero or more times
+      BASE_NAME_REGEXP = %r{[[:alpha:]](?:[\w\s]*)}x.freeze
```
misaka, line 6:
```suggestion
BASE_NAME_REGEXP = %r{[[:alpha:]](?:[\w\s]*)}.freeze
```

🤓

PR #391, opened 2022-10-04
Fix exclusive checkbox option example

c0nspiracy:
The exclusive checkbox option example has inconsistent checkbox names `countries` and `languages` which causes the live guide to have incorrect behaviour.


peteryates (approved):
Ah, well spotted! Thanks.

PR #397, opened 2022-11-04
Append custom custom classes to labels

cpjmcquillan:
Before, custom classes passed in the `label` hash would override any existing classes.

For example:

```erb
<%= form_with model: @model do |f| %>
  <%= f.govuk_text_area :details, label: { text: "Some details", class: "special-class" } %>
<% end %>
```

would output:

```html
<!-- form markup --!>

<label for="model-details-field" class="special-class">
  Some details
</label>

<!-- form markup --!>
```

Taking into consideration how the rest of the library behaves, this behvaiour was unexpected and meant users would need to explicitly provide default label classes.

This change makes use of the `HTMLAttributes` and `HTMLClasses` traits so that label elements build classes as usual —except any custom classes are appended.

For example:

```erb
<%= form_with model: @model do |f| %>
  <%= f.govuk_text_area :details, label: { text: "Some details", class: "special-class" } %>
<% end %>
```

would now output:

```html
<!-- form markup --!>

<label for="model-details-field" class="govuk-label special-class">
  Some details
</label>

<!-- form markup --!?
```

This change in behaviour will affect users of the gem who are customising label classes but _don't_ want default classes to be applied.

I'm not sure if that is something we need to worry about though, since the form builder is intended to build GOV.UK (or other brand) forms.

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/caption.rb
+++ b/lib/govuk_design_system_formbuilder/elements/caption.rb
@@ -2,6 +2,8 @@ module GOVUKDesignSystemFormBuilder
   module Elements
     class Caption < Base
       include Traits::Localisation
+      include Traits::HTMLAttributes
+      include Traits::HTMLClasses
 
       def initialize(builder, object_name, attribute_name, text: nil, size: config.default_caption_size, **kwargs)
         super(builder, object_name, attribute_name)
@@ -14,10 +16,14 @@ module GOVUKDesignSystemFormBuilder
       def html
         return unless active?
 
-        tag.span(@text, class: @size_class, **@html_attributes)
+        tag.span(@text, **attributes(@html_attributes))
       end
 
     private
+
+      def options
+        { class: @size_class }
+      end
 
       def active?
         @text.present?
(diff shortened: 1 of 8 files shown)


cpjmcquillan:
Hey @peteryates

I'm lucky enough to be back using this wonderful gem, and I found myself banging my head just now trying to figure out why all the default classes were being overriden 😆.

Is there a reason the existing behaviour was the way it was?

I'd also be interested to know how this might impact other users.

Let me know what you think, and if you need anything from me!

peteryates:
I think I just missed it 😞 

Amazed nobody's raised a bug in all that time. Will properly review and get this merged over the weekend.

Thanks for the fix, glad to have you back!

peteryates:
Perfect, thanks @cpjmcquillan. While reviewing I noticed that legends, captions and hints suffered from the same problem, so I followed up the commit with one that extends your implementation to them too.

peteryates (approved):

peteryates:
I'll try and get a release out tomorrow, should be able to find a bit of time.

PR #398, opened 2022-11-14
GOV.UK Frontend 4.4.0

peteryates:
GOV.UK Frontend [was released](https://github.com/alphagov/govuk-frontend/releases/tag/v4.4.0) earlier today.

It looks like the vast majority of the changes are in the JavaScript and don't apply to this gem.

The ones that do are in the error summary; the following suggestions have been implemented 7ee5ce140c9f3ddc6.

> * Remove `aria-labelledby="error-summary-title"` and `role="alert"` from the
>   parent element (govuk-error-summary)
> * Add a div wrapper around the contents of govuk-error-summary with the
>   attribute `role="alert"`
> * Remove `id="error-summary-title"` from the error summary h2
>   (`govuk-error-summary__title`)

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/error_summary.rb
+++ b/lib/govuk_design_system_formbuilder/elements/error_summary.rb
@@ -18,14 +18,16 @@ module GOVUKDesignSystemFormBuilder
         return unless object_has_errors?
 
         tag.div(**attributes(@html_attributes)) do
-          safe_join([title, summary])
+          tag.div(role: "alert") do
+            safe_join([title, summary])
+          end
         end
       end
 
     private
 
       def title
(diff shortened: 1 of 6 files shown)


lib/govuk_design_system_formbuilder/elements/error_summary.rb
```diff
@@ -18,14 +18,16 @@ def html
         return unless object_has_errors?
 
         tag.div(**attributes(@html_attributes)) do
-          safe_join([title, summary])
+          tag.div(role: "alert") do
+            safe_join([title, summary])
+          end
```
peteryates, line 23: > * Add a div wrapper around the contents of `govuk-error-summary` with the attribute `role="alert"`

PR #399, opened 2022-11-21
Remove unneeded `x` from localisation regexp

peteryates:
This regexp was previously split over several lines with comments explaining what was going on. It was since simplified but the x wasn't removed.

Thanks @misaka for spotting this!


misaka (approved):
🤓 🚀

PR #403, opened 2022-12-16
Remove `[0-9]*` pattern from numeric text example

peteryates:
Using the pattern attribute is no longer advised in the design system.


edwardhorsford (approved):
Nice!

PR #429, opened 2023-05-23
Add support for inverse buttons

### LINKED ISSUE #427: Support inverse button style
See https://github.com/alphagov/govuk-frontend/pull/3556. Not sure what release of `govuk-frontend` this is scheduled for.

`inverse: true`?

peteryates:
Inverse buttons [were recently added to govuk-frontend](https://github.com/alphagov/govuk-frontend/pull/3556) but haven't yet made it to the GOV.UK Design System. This change adds support for them so we're ready when they land.

- Add support for inverse buttons
- Add an example of inverse buttons to the guide

Fixes #427 

image

### DIFF
--- a/guide/content/form-elements/submit.slim
+++ b/guide/content/form-elements/submit.slim
@@ -24,6 +24,10 @@ p.govuk-body
   code: warning_button)
 
 == render('/partials/example.*',
+  caption: 'Generating submit button with inverted colours',
+  code: inverse_button)
+
+== render('/partials/example.*',
   caption: 'Generating a submit button with a custom class',
   code: custom_classes_button)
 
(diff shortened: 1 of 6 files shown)


paulrobertlloyd:
Might it be possible to show the inverse button example on a brand blue background?

peteryates:
@paulrobertlloyd which shade? 😄

paulrobertlloyd:
Unfortunately, it seems as though the inverse button style is hard coded to use `govuk-colour("blue")` and not `$govuk-brand-colour`. Maybe that’s fine. Anyway, that’s what should be used for the example’s background (`govuk-colour("blue")`).

PR #440, opened 2023-07-20
Allow merging data-aria-controls

sobakasu:
We have a situation where we want to add additional DOM ids to the `aria-controls` attribute of a govuk form element.

The `x-govuk/html-attributes-utils` is used to merge attributes. It handles merging `aria-controls` but not `data-aria-controls`. It's unclear to me why `data-aria-controls` is used here. It seems like the front end javascript moves the data-aria-controls into aria-controls. Can we just set `aria-controls` directly?

If setting `data-aria-controls` rather than `aria-controls` is important, perhaps we could instead modify `x-govuk/html-attributes-utils` to also merge `%i[data aria controls]`

### DIFF
--- a/lib/govuk_design_system_formbuilder/traits/fieldset_item.rb
+++ b/lib/govuk_design_system_formbuilder/traits/fieldset_item.rb
@@ -25,7 +25,7 @@ module GOVUKDesignSystemFormBuilder
           class: classes,
           multiple: @multiple,
           aria: { describedby: [hint_id.presence] },
-          data: { 'aria-controls' => @conditional_id }
+          data: { aria_controls: @conditional_id },
         }
       end
 


peteryates:
Thank you for your commit @sobakasu, it's very much appreciated. We set `data-aria-controls` simply because the GOV.UK Design System uses it in its examples and we try to follow them as closely as possible.

I allowed the review app for this branch to be built and the conditionally revealed examples do work just fine. I will ask the Design System team the question on why `data-aria-controls` was used and try to find out why the approach was taken.

I'll be more than happy to add it `data-aria-controls` to the list of mergeable attributes in `html-attributes-utils` too though.

peteryates:
In a follow up, I asked the team and the answer is because of progressive enhancement. Until the JavaScript module has been initialized the controlled content isn't yet controlled by the element, so if JavaScript was disabled `aria-controls` would be misleading or incorrect.

sobakasu:
> In a follow up, I asked the team and the answer is because of progressive enhancement. Until the JavaScript module has been initialized the controlled content isn't yet controlled by the element, so if JavaScript was disabled `aria-controls` would be misleading or incorrect.

ah that makes sense @peteryates . I've amended my PR to use data-aria-controls as it was before but with a symbol aria_controls instead of the string for consistency. This requires a change to the `html-attributes-utils` gem to make the specs pass. I added `%i(data aria_controls),` to `DEFAULT_MERGEABLE_ATTRIBUTES` in `HTMLAttributesUtils`.

peteryates:
Does v1.0.1 of html-attributes-utils solve the problem for you @sobakasu? It's `%i(data aria controls)` instead of `%i(data aria_controls)` but I suspect it'll work.

sobakasu:
I did try `%i[data aria controls]` but that didn't work, which is why I suggested `%i[data aria_controls]`.

`%i[data aria controls]` does this:

```ruby
<%= check_box_tag :test, 1, false, data: { hello: "1", aria: { controls: "foo" } } %>
```

```html
<input type="checkbox" name="test" id="test" value="1" data-hello="1" data-aria="{&quot;controls&quot;:&quot;foo&quot;}" />
```

(it seems that attribute names are only joined for the first level of nesting)

peteryates:
Ok, yeah that makes sense. Will fix shortly.

PR #458, opened 2023-12-27
Documentation design update

paulrobertlloyd:
Following on from https://github.com/x-govuk/govuk-components/pull/486, this PR:

- Fixes the logo in the header now that we need to support the new v5 way of doing things
- Uses the Masthead component styles from the GOVUK Prototype Components project
- Adds a modifier class to the header on the index page to account for the masthead
- Increases the overall width of the page container, and adds sub-navigation to all pages
- Moves content from the homepage onto the Getting started page (with some light editing)
- Uses the related navigation component styles from the GOVUK Prototype Components project for related information

Screenshot of Getting started page showing updated logo in the header and sub navigation with the current page selected.


peteryates (approved):
Spot on, looks great. Thank you!

PR #474, opened 2024-02-03
Remove guide hover colour override

peteryates:
The only place I can see it have an impact is on small radios and the dark blue looks a bit odd.

Here is the Design System on the left and the guide on the right before the change:

Screenshot from 2024-02-03 19-47-58

After the change they'll look the same. Not sure what else this affects though! 🙈

### DIFF
--- a/guide/content/stylesheets/application.scss
+++ b/guide/content/stylesheets/application.scss
@@ -4,8 +4,8 @@ $govuk-assets-path: "/assets/";
 $govuk-font-family: system-ui, sans-serif;
 $govuk-brand-colour: #2288aa;
 $govuk-link-colour: #006688;
-$govuk-hover-colour: #004466;
-$govuk-visited-colour: #333366;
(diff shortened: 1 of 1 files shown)


guide/content/stylesheets/application.scss
```diff
@@ -4,7 +4,6 @@ $govuk-assets-path: "/assets/";
 $govuk-font-family: system-ui, sans-serif;
 $govuk-brand-colour: #2288aa;
 $govuk-link-colour: #006688;
-$govuk-hover-colour: #004466;
 $govuk-visited-colour: #333366;
```
paulrobertlloyd, line 7: Oops, these 2 colour variables should have link in their name:

```suggestion
$govuk-link-hover-colour: #004466;
$govuk-link-visited-colour: #333366;
```

paulrobertlloyd (commented):
Good catch; added correct SCSS variables.

PR #481, opened 2024-02-22
Add password input

peteryates:
The 'password input' is a new form component added to version 5.3.0 of the GOV.UK Design System.

This change attempts to implement it with all of its custom behaviour and document it in the guide.


spec/govuk_design_system_formbuilder/builder/date_spec.rb
```diff
@@ -1,7 +1,7 @@
 describe GOVUKDesignSystemFormBuilder::FormBuilder do
   include_context 'setup builder'
 
-  describe '#date_input_group' do
+  describe '#govuk_date_field' do
```
peteryates, line 4: Spotted a historic copy/paste error 🙈

guide/lib/examples/password.rb
```diff
@@ -0,0 +1,31 @@
+module Examples
+  module Password
+    def default_password
+      <<~PASSWORD
+        = f.govuk_password_field :password, label: { text: "Enter your password" }
```
paulrobertlloyd, line 5: To be consistent with the example used in the Design System documentation.

```suggestion
        = f.govuk_password_field :password, label: { text: "Password" }
```

guide/lib/examples/password.rb
```diff
@@ -0,0 +1,31 @@
...
+      PASSWORD
+    end
+
+    def password_with_a_captioned_heading_label_and_hint
+      <<~PASSWORD
+        = f.govuk_password_field(:pin,
+            label: { text: "Enter your PIN", tag: "h2", size: "l" },
+            caption: { text: "Security", size: "m" },
+            hint: { text: "Your PIN was emailed to you when you registered for this service" })
```
paulrobertlloyd, line 14: Should probably use another example, only referring to passwords, as the Design System guidance states that this input should only be used for passwords, and not other types of security information.

guide/lib/examples/password.rb
```diff
@@ -0,0 +1,31 @@
...
+    def password_with_custom_text
+      <<~PASSWORD
+        = f.govuk_password_field(:secret_code,
+            label: { text: "Enter your secret code" },
+            show_password_text: "Display",
+            hide_password_text: "Conceal",
+            show_password_aria_label_text: "Display the secret code",
+            hide_password_aria_label_text: "Conceal the secret code",
+            password_shown_announcement_text: "The secret code has been displayed",
+            password_hidden_announcement_text: "The secret code has been concealed")
```
paulrobertlloyd, line 27: Same as above. Maybe these examples could demonstrate using these parameters to enable localisation?

guide/lib/helpers/link_helpers.rb
```diff
@@ -24,6 +24,7 @@ def navigation_links
           "Checkboxes" => "/form-elements/checkboxes/",
           "Date input" => "/form-elements/date-input/",
           "File upload" => "/form-elements/file-upload/",
+          "Password fields" => "/form-elements/password/",
           "Radios" => "/form-elements/radios/",
```
paulrobertlloyd, line 28: As previous comment, and also to be internally consistent with URL slugs:

```suggestion
          "File upload" => "/form-elements/file-upload/",
          "Password input" => "/form-elements/password-input/",
          "Radios" => "/form-elements/radios/",
```

paulrobertlloyd (dismissed):
Just did a very quick pass. Main points:

* Consistently use ‘Password input’ for the name of this field
* Don’t use examples that go against the guidance for this input

guide/content/form-elements/password.slim
```diff
@@ -0,0 +1,31 @@
+---
+title: Password fields
```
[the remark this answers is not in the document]
peteryates: I thought about that but I think consistency with Rails is more important here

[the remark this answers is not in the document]
paulrobertlloyd: Could argue the case either way, but the rest of the for builder documentation already uses ‘input’ for the page titles (‘Date input’, ‘Text input’). I think that’s the right way to go; use the name of the component in the Design System, then in the documentation, talk about fields when relating to implementation.

guide/lib/examples/password.rb
peteryates: Fair - I need unique attribute names here which is why I used PIN. I'll rename to `password_1`, `password_2` and `password_3`.

guide/content/form-elements/password.slim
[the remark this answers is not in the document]
peteryates: Addressed in 507a0d1e5605e.

frankieroberto (approved):
Code looks good to me. 👍

Any thoughts on whether this counts as a breaking change or not? On the one hand it’s a new feature, but on the other it re-uses the existing `govuk_password_field` method.

Whilst I think in most circumstances, having the show/hide button appear on existing password fields should be fine (and good) - I guess the changed markup _might_ break some tests, and possibly there might be some layouts which need adjusting to make room for the extra button? 🤔

peteryates:
Yeah I was going to cover that in the release notes. Technically breaking _but_ the old one didn't exactly do very much to start with, can't really see it causing any problems.

PR #514, opened 2024-07-21
Standalone labels

peteryates:
This PR exposes `GOVUKDesignSystemFormBuilder::Elements::Label` directly via `#govuk_label`. It's intended to allow authors to place the label somewhere other than immediately above the input.

Its usage isn't covered by the design system so comes with the risk that the error message will still be rendered next to the input which will likely break layouts when used in confined spaces (i.e., table cells)

The regular label arguments will still work so authors can customise the label's:

* size
* wrapping tag
* caption
* visibility

### DIFF
--- a/spec/govuk_design_system_formbuilder/builder/label_spec.rb
+++ b/spec/govuk_design_system_formbuilder/builder/label_spec.rb
@@ -1,0 +1,75 @@
+describe GOVUKDesignSystemFormBuilder::FormBuilder do
+  include_context 'setup builder'
+
+  describe '#govuk_password_field' do
+    let(:custom_label_text) { "What's your name?" }
+    let(:method) { :govuk_label }
+    let(:attribute) { :name }
+    let(:args) { [method, attribute] }
+    let(:kwargs) { { text: custom_label_text } }
+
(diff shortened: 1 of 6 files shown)


spec/govuk_design_system_formbuilder/builder/label_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    let(:custom_label_text) { "What's your name?" }
+    let(:method) { :govuk_label }
+    let(:attribute) { :name }
+    let(:args) { [method, attribute] }
+    let(:kwargs) { { text: custom_label_text } }
+
+    subject { builder.send(*args, **kwargs) }
+
+    specify 'renders a label with the right custom text' do
+      expect(subject).to have_tag('label', text: custom_label_text, with: { class: 'govuk-label' })
```
frankieroberto, line 14: Worth adding a test for the `for` attribute too, given that this will be needed to link the label to the input?

peteryates: Missed this, but yeah it is. Will add it.

PR #515, opened 2024-07-30
Allow brand overrides on a per-helper basis

peteryates:
Allow brand to be overridden on a per field basis.

This feature was originally requested against [govuk-components](https://github.com/x-govuk/govuk-components/issues/547) but I'm testing the approach here first - it should be easy to reuse there.


peteryates:
Closing for now, will reinvestigate later if needed.

PR #538, opened 2024-11-24
Use `Struct` in examples

tvararu:
`view_component` v3.20.0 has removed `require "ostruct"` from their codebase:

https://github.com/ViewComponent/view_component/releases/tag/v3.20.0

As such, users that want to use `OpenStruct` have to add the `require` call to their codebases.

Instead, we can use `Struct`, which works similarly and doesn't have to be required.


peteryates (approved):
Looks great, thank you ♥️

PR #561, opened 2025-05-23
Make hardcoded text translatable

peteryates:
There are some bits of text in the library that are hardcoded and there's no ability to override/configure them.

This PR intends to fix them which will make it easier to use the library in non-English or multilingual applications.


chao-xian:
TYSM! 🫶

PR #599, opened 2026-01-23
Add `govuk_time_field`

thomasleese:
This adds a new type of field which can be used to render a time input. It works in a very similar way to the date field, but instead of day, month and year the user is asked for hour, minute and second.

Ruby doesn't have a built-in type capable of modelling just the time component of a date/time. Instead, the value is stored as a `Time` object where the day, month and year are ignored. When Rails is dealing with the PostgreSQL `TIME` type, the date component is set to 2020-01-01.

To implement this I've introduced a `DateInput` trait which is shared across both the date and the time fields. Any common logic across the two fields is in the trait.

I can't see a time field in the GOV.UK Design System so I've added a warning to the documentation that this is not a standard component.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,8 @@ AllCops:
     - govuk_design_system_formbuilder.gemspec
 Rails/Date:
   Enabled: false
+Rails/TimeZone:
+  Enabled: false
 Rails/ActiveSupportAliases:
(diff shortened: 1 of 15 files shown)


peteryates (approved):
Looks great. Thank you so much. This will be part of the upcoming 6.0.0 release.

PR #619, opened 2026-04-07
Add proc support to govuk_error_summary

m-dango:
While using the form builder, we found that we were unable to translate the default error summary heading in the same way we had done for the submit button e.g. `-> { I18n.t('translation') }`.

This change allows the error summary to call procs in the same way the submit element does.

Additionally, it adds a test case for proc support on the submit button.


m-dango:
I need to delete the forked repo for reasons outside of my control, so this may get reopened in another PR if that impacts this.

PR #628, opened 2026-06-11
Fix #627: Merge password form group options

### LINKED ISSUE #627: Password input does not support passing class or data attributes via form_group
My understanding is that inputs are expected to be able to pass HTML attributes to form group by passing a hash to `form_group:`. The Password input does not support class and data attributes because it specifically overwrites them:

```ruby
def form_group_options
  {
    **@form_group,
    **i18n_data,
    class: %(#{brand}-password-input),
    data: { module: %(#{brand}-password-input) },
  }
end
```

Given this example:

```ruby
form.govuk_password_field(..., form_group: { class: "example", data: { controller: "password" } }, ...)
```

**Actual behaviour**

```html
<div class="govuk-form-group govuk-password-input" data-module="govuk-password-input">
```

**Expected behaviour**

```html
<div class="govuk-form-group govuk-password-input example" data-module="govuk-password-input" data-controller="password">
```

sfnelson:
Use `HTMLAttributesUtils` to merge `form_group` options passed to `govuk_password_input` instead of clobbering them.

### DIFF
--- a/lib/govuk_design_system_formbuilder/elements/password.rb
+++ b/lib/govuk_design_system_formbuilder/elements/password.rb
@@ -2,6 +2,7 @@ module GOVUKDesignSystemFormBuilder
   module Elements
     class Password < Base
       using PrefixableArray
+      using HTMLAttributesUtils
 
       include Traits::Error
       include Traits::Hint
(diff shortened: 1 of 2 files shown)


peteryates:
Oh I'm sorry, I missed this PR entirely but saw and fixed the issue earlier in #629

PR #638, opened 2026-07-14
Fix password input error class

m-dango:
When using the form builder, I noticed that the password field was not being highlighted in red on error. This appears to be caused by the error class appended to the password field being incorrect.

The error message in the [design system](https://design-system.service.gov.uk/components/password-input/#error-messages) shows the example to just have `govuk-input--error` rather than `govuk-password-input--error`.


peteryates:
Thank you @m-dango. Didn't realise this component deviated from the convention!

